PR #23, opened 2012-12-19
Fix prototype of SourceTransformer caller

blanchonvincent:


lisachenko:
Build is failed ) Need to fix it before merge

PR #24, opened 2012-12-19
Proposition to improve the transformer workflow

blanchonvincent:
Wrap the source with the stream metadatas. Because, someone transformer
need source, someone do not need source. With a global wrapper (the
StreamMetaData), each transform is based on only one object.


blanchonvincent:
@lisachenko it's just a proposition, give me your feedback.

lisachenko:
@blanchonvincent good idea! :+1: This gives more stable interface for transformers. 

However, for future I want to use PHP-parser library and Abstract Syntax Tree as source instead of string and StreamMetaData. Transformers will be rewritten as node visitors, this will give more control on source code.

PR #39, opened 2013-01-27
Refactoring: use WeavingTransformer instead of AopProxyTransformer

Wolg:


lisachenko:
Thanks) Source code in master branch can be fragile without tests. Still want to write tests) But haven' time for this.

PR #40, opened 2013-03-03
Implement pointcut parser engine

lisachenko:
This PR introduces new feature: Pointcut parser engine. Go! library now has a lexer, grammar and parser that allows to generate AST tree. 
Many classes were refactored, some of unused class were deleted, some of them moved to the Pointcut namespace.


lisachenko:
@cursedcoder Can you mark the blank line in the concrete commit?

For 'use' statements I have my own rule: first block is system classes, next block is separated with blank line and defines project classes and the last one is vendor classes.

PR #72, opened 2013-07-07
Faster debug mode with cache on

mpoiriert:
This will check if the file need to be recompile on various criteria before doing it.

**Wasn't able to run unit test since they where failling before the change** Test have been updated for this change and code have been tested with a application in development.

In my project in debug mode speed have pass from about 25 seconds loading a page to less than 2 seconds who's the normal behavior in dev...


lisachenko:
Close this PR because CachingTransformer should optimize caching of modified files for faster debug.

PR #78, opened 2013-07-30
fix when include used with ternary operator

DavertMik:
fix for "$xx = ? include(xx) : 'yy'" thing (Yii2)

fix when include used with ternary operator like: https://github.com/yiisoft/yii2-framework/blob/master/yii/helpers/SecurityBase.php#L123

Yep, Go AOP was patching easier then Yii2

### DIFF
--- a/src/Go/Instrument/Transformer/FilterInjectorTransformer.php
+++ b/src/Go/Instrument/Transformer/FilterInjectorTransformer.php
@@ -151,7 +151,7 @@ class FilterInjectorTransformer implements SourceTransformer
         $transformedSource = '';
         $isWaitingEnd      = false;
         foreach ($tokenStream as $token) {
-            if ($isWaitingEnd && ($token === ';' || $token === ',')) {
+            if ($isWaitingEnd && ($token === ';' || $token === ',' || $token === ':')) {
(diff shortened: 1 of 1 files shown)


DavertMik:
Does this expression work in current version? 
Is that used somewhere?
Should `?` also be included as ending token?

DavertMik:
It's would be great to solve this in any way, because I'm trying to write the post about Yii2 for almost a week now :(
Without this patch Yii won't work.

lisachenko:
> Does this expression work in current version? 

Yes, it works

> Is that used somewhere?

Not know exactly, but I have a feeling that I saw this expression early, so this is the case.

> Should ? also be included as ending token?

No, this will make things much worse...

Unfortunately, there is no reliable way to cover all the cases, because expression can be any valid expression with ternary operators, brackets, parenthesis, etc. It's pitty, but Go! will support only restricted sets of `include` expression. This is known limitation.

DavertMik:
Oh, ok, thanks anyway.
I will send alternative PR into Yii2 core.

DavertMik:
Done :)
https://github.com/yiisoft/yii2/pull/697

Issue #71, opened 2013-07-07
Error with class/file that have specific instruction at the end (ex: Propel1 model classes)


mpoiriert:
I'm using propel 1 in a project and I get a issue with the generated file.

At the end of the file propel call a static method on the current declared class. Since the class with the good name is appended to the class it cannot be found.

Here is a exemple of before and after:

```
//Before
class BaseFilePeer
{
}

BaseFilePeer::buildTableMap();


//After
class BaseFilePeer__AopProxied
{
}

BaseFilePeer::buildTableMap();

abstract class BaseFilePeer extends BaseItemPeer__AopProxied implements \Go\Aop\Proxy
{
}
```

I've tried modifying the WeavingTransformer to append the child class just after the first class but the ReflectionElement just give access to the last line of the class and not the 'char' position in the file (it's probably why you don't support 2 namespaces in files).

Do you have any suggestion because this is a big issue for my project...

lisachenko:
Hi, it seems that the Propel violates PSR-1. My framework currently doesn't support such files. Can you add this directory to the ignore list for AOP? Or you need to use AOP for Propel models?

mpoiriert:
I might use aop for the propel model but other thant that:
1. Many folders contain propel model, and external libraries might too
2. Folders also contain class that are not propel model...

Does make it harder to do a clean exclusion...

lisachenko:
Yeap...Houston, we have a problem...

lisachenko:
Nice bug report ) Thank you very much! 
BTW, how can I easily reproduce this issue? I'll try to found a way to solve it.

mpoiriert:
From what I see juste creating a class with a static method that is call at the end of the file should trigger this issue.

mpoiriert:
FYI, I'm doing my masters project by creating a framework that I can be use within other framework by creating bridge. I'm using your AOP implentation in the core of the framework. Other than regular usage of AOP, I'll be using it to create bridge for the other framework to be the less intrusive possible. I'm using it on a concreate project at my job... I'll probably open real use case issue often, I'll try to do patch when I can but at least I'll document the issues as much as possible...

lisachenko:
Very interesting project! I'm sure that it will be awesome. Thank you one more time for so documented issues, I will try to fix them asap.
Unfortunately, I haven't much time for my project (it's just a hobbie) and can work only at weekends, so excuse me for any delays.

worried-networking:
[...]

PR #83, opened 2013-08-13
Use tokenizer for inserting child class

worried-networking:
With this change we can work with AspectMock & Yii.

Also, constructions like this:

``` php
<?php

class Classname {
    public function autoload($class) {
        include $class . '.php';
    }
}

spl_autoload_register(array('Classname', 'autoload'));
```

now processed successfully (before that patch we get a fatal error - because `Classname` was replaced with `Classname__AopProxied` and `Classname` definition appended to the end of file, after `spl_autoload_register`)

### DIFF
--- a/src/Go/Instrument/Transformer/WeavingTransformer.php
+++ b/src/Go/Instrument/Transformer/WeavingTransformer.php
@@ -108,6 +108,7 @@ class WeavingTransformer extends BaseSourceTransformer
         /** @var $namespaces ParsedFileNamespace[] */
         $namespaces = $parsedSource->getNamespaces();
         assert('count($namespaces) < 2; /* Only one namespace per file is supported */');
+        $lineOffset = 0;
 
         foreach ($namespaces as $namespace) {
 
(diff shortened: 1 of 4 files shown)


worried-networking:
And this PR fixes #71

lisachenko:
Hello! Thank you very much for this PR, it's really good. Can I suggest to refactor this PR to use already available information about last position and last line of the class instead of tokenizer? Before line 147 in WeavingTransformer:

``` php
$lastLine = $class->getEndLine(); // returns the last line of class
$lastPosition = $class->getEndPosition(); // returns the last position
/// ...
$metadata->source = substr($metadata->source, 0, $lastPosition) . $child . substr($metadata->source, $lastPosition+1);
```

I'm not tested this, but this should be much faster than analyzing tokens. If you could attach a unit test for this it will be great

worried-networking:
Your suggestion is great. Unfortunately I have no success with `$class->getEndPosition()` - looks like it returns position without comments. So I just made a little magic with `$class->getEndLine()` and `explode`.
Unit-test is also added.

PR #85, opened 2013-08-14
Add a support for "not so simple" includes

worried-networking:
E.g. with extra function call, like "include($this->getViewFile($view,$data['code']));"


lisachenko:
Well done!

PR #86, opened 2013-08-14
Allow dynamic arguments count in ClassProxy

worried-networking:


lisachenko:
Thank you for this PR. However, I want to discuss another ways of handling dynamic arguments. Your patch will break the passing by reference feature for methods as func_get_args() doesn't handle references correctly. This is a minor issue for this PR. Major issue is performance. AOP should be as fast as possible, so it was hard to implement  the fastest method of join point invocation. Calling func_get_args() will slow slow down the speed of invocation for several times.
My first idea about dynamic count was to add special "@Dynamic" annotation to the method docBlock. Then ClassProxy can check this annotation with strpos() and then add missing parameters with func_get_args(): $args = array($a, &$b, $c = null) + func_get_args(). In this case overhead will be only for the methods with dynamic arguments. Or we can have a special method to declare signatures of methods by hand. This will be the best way to add dynamic arguments for method because there won't be any extra checks in the ClassProxy, only fastest isset($this->dynamicMethods[$class.$method]) check.

worried-networking:
Made a little rework, what do you think about this variant?

worried-networking:
And yes, I forgot to add/fix unit-tests :( I'll do it later.

worried-networking:
Ok, added some tests

worried-networking:
Кстати, Александр, после этого PR наконец-то можно будет писать функциональные тесты для Yii1 на Codeception + AspectMock. @DavertMik будет рад, я думаю :) Unit-тесты заработали уже после #83.

Anyway - few days later I'll return to `FilterInjectionTransformer` - I want to add support for expressions like #78.

worried-networking:
@lisachenko do you have any comments for latest implementation? I'm looking forward to see this feature in master :)

lisachenko:
@zvirusz PR is nice and I can merge it after fixing the small notice: do not add an additional empty `array()` to the list of arguments for invocation if a method doesn't expect any parameters. Generated code should be clean and shouldn't contain unnecessary arguments, as this will increase parse/invocation time.

lisachenko:
PS. Please, squash a commits before the merge to remove an irrelevant changes in the tests) Thanks!

worried-networking:
Done

lisachenko:
Accepted

PR #87, opened 2013-08-15
Add support for ternary operator in includes

worried-networking:


lisachenko:
Nice patch )
However, I want to give your a notice that Go! shouldn't care about all possible language constructions because it's too complicated task. By adding more logic into the parsing we will slow down the overall performance. Please, do not add more logic in the future into the FilterInjectorTransformer to handle all possible constructions.

BTW, this patch is too tricky and I want to test it at weekend before merge. Be patient )

worried-networking:
> Please, do not add more logic in the future into the FilterInjectorTransformer to handle all possible constructions.

Ok, I'll keep in mind.

> BTW, this patch is too tricky and I want to test it at weekend before merge. Be patient )

I don't worry about this PR, but I want to see #86 in master ASAP :)

> Please, replace in_array() check with explicit checks. This should be faster.

Yep, it's faster. But only for 0.1E-8 :) Changed.

> Oh, can you make this more clear?

Ok, I'll rework it on weekends.

worried-networking:
Is it enough to support only this 2 expressions?

``` php
$x = $var ? include $y : $z;
include $var ? $y : $v;
```

If yes - this PR can be much more simpler :)

lisachenko:
> Yep, it's faster. But only for 0.1E-8 :) Changed.

Yeah, it's an AOP. Everything in this projects should be optimized, because each file with include can contain thousands of tokens, each application can have many files with includes, so there will be about 1e8-1e9 iterations with checks for huge application. This small improvement can result in 10-1000ms performance gain )

> Is it enough to support only this 2 expressions?
> 
> ``` php
> $x = $var ? include $y : $z;
> include $var ? $y : $v;
> ```

I think that it will be enough, but if you be able to make a clean patch for current PR with fastest checks I can change my mind )

worried-networking:
Ok, reworked, can you take a look?

lisachenko:
Accepted )

PR #92, opened 2013-08-20
Unset foreach value which passed by reference

worried-networking:


lisachenko:
Very nice! Thank you )

PR #93, opened 2013-08-20
Do not change autoloaders' order on AopComposerLoader initialization

### LINKED ISSUE #91: GO! AOP breaks composer autoloader
I'm trying to work with `PHPExcel` and at some point autoloading fallbacks to Yii's built-in and then fails. At this point I don't need GO! Aop features so I've commented out `AopComposerLoader.php` in composer's `autoload_files.php` - and everything works ok.

@lisachenko, if you don't have any ideas 'why?' - I'll create sample failing app in next few days.

worried-networking:
Actually I didn't fixed #91 by #92 - it's just changed behavior of this error. And this PR makes my test-code working completely correct.

### DIFF
--- a/src/Go/Instrument/ClassLoading/AopComposerLoader.php
+++ b/src/Go/Instrument/ClassLoading/AopComposerLoader.php
@@ -60,6 +60,7 @@ class AopComposerLoader
         $loaders = spl_autoload_functions();
 
         foreach ($loaders as &$loader) {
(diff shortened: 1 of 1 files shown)


worried-networking:
Arghr... It's fixed one issue and made few our tests fails :(
Ignore this PR for a while, please.

worried-networking:
@lisachenko ok, fixed. Can you take a look?

PR #103, opened 2013-12-01
Warmup

mpoiriert:
This will do a warmup base on the map file if the dump-autoload --optimize have been in composer. A method is also available to pass a array of files if needed.

The unit test have been started but are not really efficient since we should check if the proper file have been generated. But since I do not no how to reproduce the All the container configuration in a test environment it's pretty hard to do...

I've try to modify the minimum the actual class, so just one method to access the initial Autoloader have been add to existing class.

In my project I have more than 4k classe to generate, this can take a lot of time and up to 2G of memory (I didn't look for where memory is kept in the system...)

### DIFF
--- a/src/Go/Instrument/CacheWarmer.php
+++ b/src/Go/Instrument/CacheWarmer.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * Created by JetBrains PhpStorm.
+ * User: Martin
+ * Date: 13-12-01
+ * Time: 11:30
+ * To change this template use File | Settings | File Templates.
+ */
+
+namespace Go\Instrument;
+
+
+use Go\Instrument\ClassLoading\AopComposerLoader;
+use Go\Instrument\Transformer\FilterInjectorTransformer;
+
+class CacheWarmer
+{
+    public function warmUp()
+    {
+        foreach(spl_autoload_functions() as $autoLoader) {
+            if(!is_array($autoLoader)) {
+                continue;
+            }
+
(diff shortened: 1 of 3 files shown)


lisachenko:
Thank your for this PR ) I have some ideas for improvement before accepting this patch. 
1. In my opinion it isn't correct to load only class map from the composer. Cache warmer should receive a configuration (options) from the kernel instead and use an iterator object for traversing all files. This iterator should be based on information from aop kernel, such as application directory, cache path, include paths and exclude paths.
2. I have a suggestion to register this command as a binary for composer package and put this shell script into `bin` directory. This shell script should use symfony/console component and specify it as an optional requirement. To have an idea, look at https://github.com/jakubledl/dissect/blob/develop/composer.json#L18-L26

lisachenko:
I have implemented more powerful command for warmup that is based on symfony/console component. Could you try it for your project? With composer this can be tested with `composer require lisachenko/go-aop-php:dev-feature/cache-warmup` and then just run `./vendor/bin/warmup` with a path to bootstrap file with aspect kernel configuration.

lisachenko:
See PR #104 for details. I also added a CleanableMemory backend for token reflection that should help with high memory consumption and fatal errors during proxies compilation.

mpoiriert:
I'm working for company that are doing christmas product so I'll probably
check that at the begining of january (I'm kind in the rush now....).

lisachenko:
Oh, it's ok. I'm busy too with a launch of new services. Wish you a success with Christmas holidays!

PR #121, opened 2014-01-26
Update file headers to use more friendly name of project

lisachenko:


composer.json
```diff
@@ -25,7 +25,7 @@
     "bin": ["bin/warmup"],
     "authors": [
         {
-            "name": "Lissachenko Alexander",
+            "name": "Lisachenko Alexander",
```
cordoval, line 28: i guess is a transliterated form, i never knew

lisachenko, line 28: Oh, it's a funny fact. Russian "s" (Ли **с** аченко) sound is softer than an equivalent single "s" sound in English. So, putting a double "s" makes sure that English speaker pronounce my last name correctly. But I want to use transliteration, because it's more natural now (passport, credit cards, etc).

PR #134, opened 2014-03-11
defensive try..catch block for type hint bug of Intl ext in PHP 5.5.x

hax:
Current demo of interception for internal functions will throw exception in PHP5.5 . See https://gist.github.com/hax/9479075 for the root cause.

### DIFF
--- a/src/Go/Proxy/FunctionProxy.php
+++ b/src/Go/Proxy/FunctionProxy.php
@@ -297,8 +297,15 @@ BODY;
         $type = '';
         if ($parameter->isArray()) {
             $type = 'array';
-        } elseif ($parameter->getClass()) {
-            $type = '\\' . $parameter->getClass()->name;
+        } else {
+            try {
+                $cls = $parameter->getClass();
+                if ($cls) {
+                    $type = '\\' . $cls->name;
+                }
+            } catch (\ReflectionException $e) {
+                // ignore
+            }
         }
         $defaultValue = null;
         $isDefaultValueAvailable = $parameter->isDefaultValueAvailable();


lisachenko:
I think that this try..catch block is not really needed, it will be better to submit a bug on PHP Intl itself. Also, you can avoid this exception by excluding this function from pointcut.
Please, could you give me you feedback about your usecase for intercepting the execution of that function?

lisachenko:
@hax ping, I gonna close this issue as it's not related to the framework itself.

hax:
Sorry I'm too busy these days...

Yes there will be no usecase to intercept `intltz_from_date_time_zone`, but the problem here is not about whether we could intercept it, the problem is it's too crazy to ask devs excluding it for every function interception.

And I know it's a bug of PHP Intl itself, we will suffer this problem until they fix it. It's hard for most developers to know the gotcha, all they will see is just a meanless exception.

Please note that Intl extension is bundled with PHP by default, that's why I send this pull request as a workaround.

lisachenko:
@hax ok, I'll decide later, could you please report this bug on bugs.php.net and add a link here?

hax:
This is my first time to report bug of PHP... :)

https://bugs.php.net/bug.php?id=66921

lisachenko:
@hax very nice, we can wait a response from PHP developers about that bug

lisachenko:
I'm decided not to include this try..catch block. Hope, that this bug in PHP will be fixed sometime in the future.

hax:
This bug has been fixed 29 hours ago :) http://git.php.net/?p=php-src.git;a=commit;h=8ba240a70b22f50483818cc0302fd264b1a22fc9

So the next release of PHP (5.5.15) should be ok.

lisachenko:
Oh, great! ) Congrats! Your first bug in PHP was fixed.

PR #192, opened 2014-12-28
Fixed proxy creation path.

andy-shea:
Uses the original reflection class's filename as base for proxy class's path as not all namespaces correlate directly to filesystem path.  This ensures the correct path will be returned by the `MagicConstantTransformer` when calling `ReflectionClass->getFileName()`.

References #193

### DIFF
--- a/src/Instrument/Transformer/WeavingTransformer.php
+++ b/src/Instrument/Transformer/WeavingTransformer.php
@@ -288,7 +288,10 @@ class WeavingTransformer extends BaseSourceTransformer
         }
         $cacheDirSuffix = '/_proxies/';
         $cacheDir       = $this->options['cacheDir'] . $cacheDirSuffix;
-        $fileName       = str_replace('\\', '/', $class->getName()) . '.php';
+        $fileName       = str_replace(
+            array('\\', $this->options['appDir'] . DIRECTORY_SEPARATOR),
+            array(DIRECTORY_SEPARATOR, ''),
+            $class->getFileName());
 
         $proxyFileName = $cacheDir . $fileName;
         $dirname       = dirname($proxyFileName);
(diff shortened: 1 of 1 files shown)


lisachenko:
Hello! Thank you for the PR!

Before merging, could you please elaborate more on why this change is needed. Maybe, add a test for your case? Or alternatively, describe with an example what is expected and what is actual.

andy-shea:
Sure.  Take this PSR-4 namespace mapping defined in a composer.json autoload block: 

```
"autoload": {
    "psr-4": {
      "My\\Namespace\\": "path/to/files"
    }
  }
```

This would mean a class with a fully qualified name of "My\Namespace\Foo\Bar\Baz" needs to be located at "path/to/files/Foo/Bar/Baz.php" relative to the composer.json dir.  Currently, the code directly maps namespace to filesystem path so the file would need to exist at "My/Namespace/Foo/Bar/Baz.php" for the `ReflectionClass->getFileName()` call to be correct but as you can see, this is far from where the file actually exists.

lisachenko:
@andy-shea thank you, it's really an issue for PSR-4, but I have a doubts that your patch will work for traditional PSR-0 loaders too. Tests are needed, because it's no so transparent.

lisachenko:
Updated your comment to point to the issue #193

andy-shea:
Actually running a quick test it has fixed the `ReflectionClass->getFileName()` calls for PSR-0 loaded libraries installed via composer as well.  I'm not sure what use cases you are testing against but the current code is incorrect for PSR-4 **and** PSR-0 libraries managed by composer.  The proposed fix is simple and robust as it merely crops the original class's filename to one that is relative to the app root.  The proxy code's directory structure will now mirror the proxied class's from the app root down. No dealing with namespaces at all.  I need to push through an update to the proposed fix as it has irrelevant remnants from a previous version (which may be why you had a feeling it would be broken for PSR-0).  The final code is simply:

```
$fileName       = str_replace($this->options['appDir'] . DIRECTORY_SEPARATOR, '', $class->getFileName());
```

lisachenko:
@andy-shea now I agree with this patch. Thank you! Will merge into 0.x and master as well.

lisachenko:
Rebased and merged into 0.x, master in https://github.com/lisachenko/go-aop-php/commit/2ac162cf6277f5e93b61d91ea1349d7a3f459729

lisachenko:
Thank you for the fix! :+1:

PR #209, opened 2015-03-13
Flexibility improvements

k-meister:
- removed code duplication & extracted real-to-cache code mapping,
- added an option that allows to set up cache file permissions,
- added a way to customize source transformers

### DIFF
--- a/src/Core/AspectKernel.php
+++ b/src/Core/AspectKernel.php
@@ -59,25 +59,11 @@ abstract class AspectKernel
     protected static $containerClass = 'Go\Core\GoAspectContainer';
 
     /**
-     * Default class name for cache path resolver, can be redefined in children
-     *
-     * @var string
-     */
-    protected static $cachePathResolverClass = 'Go\Instrument\ClassLoading\CachePathResolver';
-
-    /**
      * Aspect container instance
      *
      * @var null|AspectContainer
      */
     protected $container = null;
-
-    /**
-     * Cache path resolver instance
-     *
-     * @var null|CachePathResolver
-     */
-    protected $cachePathResolver = null;
 
     /**
      * Protected constructor is used to prevent direct creation, but allows customization if needed
@@ -107,8 +93,6 @@ abstract class AspectKernel
     {
         $this->options = $this->normalizeOptions($options);
         define('AOP_CACHE_DIR', $this->options['cacheDir']);
-
-        $this->cachePathResolver = new $this->options['cachePathResolverClass']($this);
 
         /** @var $container AspectContainer */
         $container = $this->container = new $this->options['containerClass'];
@@ -141,16 +125,6 @@ abstract class AspectKernel
     public function getContainer()
     {
         return $this->container;
-    }
-
-    /**
-     * Returns an instance of cache path resolver
-     *
-     * @return null|CachePathResolver
-     */
-    public function getCachePathResolver ()
-    {
-        return $this->cachePathResolver;
     }
 
     /**
@@ -203,7 +177,7 @@ abstract class AspectKernel
      *   debug    - boolean Determines whether or not kernel is in debug mode
      *   appDir   - string Path to the application root directory.
      *   cacheDir - string Path to the cache directory where compiled classes will be stored
-     *   cachePerms - integer Binary mask of permission bits that is set to cache files
+     *   cacheFileMode - integer Binary mask of permission bits that is set to cache files
      *   features - integer Binary mask of features
      *   includePaths - array Whitelist of directories where aspects should be applied. Empty for everywhere.
      *   excludePaths - array Blacklist of directories or files where aspects shouldn't be applied.
@@ -215,16 +189,15 @@ abstract class AspectKernel
         $features = static::getDefaultFeatures();
 
         return array(
-            'debug'     => false,
-            'appDir'    => __DIR__ . '/../../../../../../',
-            'cacheDir'  => null,
-            'cachePerms'=> null,
-            'features' => $features,
+            'debug'                  => false,
+            'appDir'                 => __DIR__ . '/../../../../../../',
+            'cacheDir'               => null,
+            'cacheFileMode'          => null,
+            'features'               => $features,
 
             'includePaths'           => array(),
             'excludePaths'           => array(),
             'containerClass'         => static::$containerClass,
-            'cachePathResolverClass' => static::$cachePathResolverClass,
         );
     }
 
--- a/src/Core/AspectKernel.php
+++ b/src/Core/AspectKernel.php
@@ -257,24 +257,12 @@ abstract class AspectKernel
                 $transformers[] = new ConstructorExecutionTransformer();
             }
 
-            return $aspectKernel->customizeTransformers($transformers);
+            return $transformers;
         };
 
         return array(
             new CachingTransformer($this, $sourceTransformers)
         );
-    }
-
-    /**
-     * An opportunity to add/remove some transformers depending on project needs
-     *
-     * @param array|SourceTransformer[] $defaultTransformers
-     *
-     * @return array|\Go\Instrument\Transformer\SourceTransformer[]
-     */
-    protected function customizeTransformers (array $defaultTransformers)
-    {
-        return $defaultTransformers;
     }
 
     /**
--- a/src/Core/AspectKernel.php
+++ b/src/Core/AspectKernel.php
@@ -12,6 +12,7 @@ namespace Go\Core;
 
 use Go\Aop\Features;
 use Go\Instrument\ClassLoading\AopComposerLoader;
+use Go\Instrument\ClassLoading\CachePathResolver;
 use Go\Instrument\ClassLoading\SourceTransformingLoader;
 use Go\Instrument\CleanableMemory;
 use Go\Instrument\PathResolver;
@@ -58,11 +59,25 @@ abstract class AspectKernel
     protected static $containerClass = 'Go\Core\GoAspectContainer';
 
     /**
+     * Default class name for cache path resolver, can be redefined in children
+     *
+     * @var string
+     */
+    protected static $cachePathResolverClass = 'Go\Instrument\ClassLoading\CachePathResolver';
+
+    /**
      * Aspect container instance
      *
      * @var null|AspectContainer
      */
     protected $container = null;
+
+    /**
+     * Cache path resolver instance
+     *
+     * @var null|CachePathResolver
+     */
+    protected $cachePathResolver = null;
 
     /**
      * Protected constructor is used to prevent direct creation, but allows customization if needed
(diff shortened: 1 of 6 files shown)


src/Core/AspectKernel.php
```diff
@@ -263,6 +293,18 @@ protected function registerTransformers()
     }
 
     /**
+     * An opportunity to add/remove some transformers depending on project needs
+     *
+     * @param array|SourceTransformer[] $defaultTransformers
+     *
+     * @return array|\Go\Instrument\Transformer\SourceTransformer[]
+     */
+    protected function customizeTransformers (array $defaultTransformers)
```
lisachenko: This method is not needed, if you want to customize the list of transformers, just override the `registerTransformers()` method. AspectMock has an example: https://github.com/Codeception/AspectMock/blob/master/src/AspectMock/Kernel.php#L62-L68

src/Core/AspectKernel.php
```diff
@@ -58,6 +59,13 @@
     protected static $containerClass = 'Go\Core\GoAspectContainer';
 
     /**
+     * Default class name for cache path resolver, can be redefined in children
+     *
+     * @var string
+     */
+    protected static $cachePathResolverClass = 'Go\Instrument\ClassLoading\CachePathResolver';
```
lisachenko: It will be better to use DI here, `AspectKernel` shouldn't know anything about classes or anything else. So, please, remove all related stuff to the `GoAspectContainer` and make a definition there.

src/Core/AspectKernel.php
```diff
@@ -65,6 +73,13 @@
     protected $container = null;
 
     /**
+     * Cache path resolver instance
+     *
+     * @var null|CachePathResolver
+     */
+    protected $cachePathResolver = null;
```
lisachenko: See my previous comment about the usage of container for storing instances. DI is the best way to decouple the source code.

src/Core/AspectKernel.php
```diff
@@ -127,6 +144,16 @@ public function getContainer()
     }
 
     /**
+     * Returns an instance of cache path resolver
+     *
+     * @return null|CachePathResolver
+     */
+    public function getCachePathResolver ()
```
lisachenko: This method violates SRP of `AspectKernel`. Should be removed. If some code wants this instance, please inject it directly via Dependency Injection principle in the container.

src/Core/AspectKernel.php
```diff
@@ -190,11 +218,13 @@ protected function getDefaultOptions()
...
+            'cachePerms'=> null,
             'features' => $features,
 
-            'includePaths'       => array(),
-            'excludePaths'       => array(),
-            'containerClass'     => static::$containerClass,
+            'includePaths'           => array(),
+            'excludePaths'           => array(),
+            'containerClass'         => static::$containerClass,
+            'cachePathResolverClass' => static::$cachePathResolverClass,
```
lisachenko: This options is not needed. Customization is possible via container class extension. You can define a parameter in the container class with default value, then extend this class in your application and configure a kernel to your specific container class with additional services/overridden parameters.

src/Core/AspectKernel.php
```diff
@@ -254,7 +284,7 @@ protected function registerTransformers()
                 $transformers[] = new ConstructorExecutionTransformer();
             }
 
-            return $transformers;
+            return $aspectKernel->customizeTransformers($transformers);
```
lisachenko: Method `customizeTransformers()` will be removed because it's no needed. Just override this method itself in your application to customize the list of extensions.

src/Instrument/ClassLoading/CachePathResolver.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+namespace Go\Instrument\ClassLoading;
+use Go\Core\AspectKernel;
+
+/**
+ * Class that manages real-code to cached-code paths mapping.
+ * Can be extended to get a more sophisticated real-to-cached code mapping
+ */
+class CachePathResolver
```
lisachenko, line 18: Could you please provide some examples how it will be used? Why do you need to change the default behaviour?

src/Instrument/Transformer/CachingTransformer.php
```diff
@@ -87,10 +96,11 @@ public function transform(StreamMetaData $metadata)
         }
 
         $originalUri = $metadata->uri;
-        $cacheUri    = str_replace($this->rootPath, $this->cachePath, $originalUri);
+        $cacheUri    = $this->kernel->getCachePathResolver()->getCachePathForResource($originalUri);
 
         $lastModified  = filemtime($originalUri);
-        $cacheModified = file_exists($cacheUri) ? filemtime($cacheUri) : 0;
+        $cacheExisted = file_exists($cacheUri);
```
lisachenko: Equal symbol is not aligned. There is also typo in the variable name, should be `$cacheExists`

src/Instrument/Transformer/CachingTransformer.php
```diff
@@ -99,6 +109,9 @@ public function transform(StreamMetaData $metadata)
             }
             $this->processTransformers($metadata);
             file_put_contents($cacheUri, $metadata->source);
+            if (!$cacheExisted && $this->cachePerms){
+                chmod($cacheUri, $this->cachePerms);
```
lisachenko: No need to check `$this->cachePerms`, just always change filemode to the specific. Suggestion: rename `cachePerms` to the `cacheFileMode`

src/Core/AspectKernel.php
k-meister: Ok, thank you,
I was thinking of doing so, but found this way of setting up container itself, so decided to mirror it.

src/Core/AspectKernel.php
k-meister: I was thinking of application code not taking care of preparing the "sandwich" of transformers correctly: application sets up linear transformations, and they are all wrapped up in cacher correctly by the framework.
I found it handy in my case. Anyway,  I'll remove the method, if it is crucial from your standpoint

src/Instrument/ClassLoading/CachePathResolver.php
k-meister, line 18: I'm trying to introduce aop support in an application that is 10+ years old :)
Its filesystem span isn't quite suitable for plain str_replacing of app_dir to cache_dir, and the only directory that contains both application and libs' source code is '/', which is impossible to use with this str_replace

src/Instrument/Transformer/CachingTransformer.php
k-meister: Ok, thank you, will fix the alignment.
As for the name - it reflects that cache file existed before file_put_contents call, should I rename it anyway?

src/Instrument/Transformer/CachingTransformer.php
k-meister: It isn't always possible to change the permissions of file that existed before file_put_contents call, since one must be the owner of the file to do so.

If we have different parts of the app running under different users, we'll get a permission problems here.

src/Core/AspectKernel.php
lisachenko: I understood your vision, but your method doesn't add significant improvements in handling of custom extensions, compared to the initial way, so I decided to reject it :)

src/Instrument/ClassLoading/CachePathResolver.php
lisachenko, line 18: Oh, man :) I hope, that there are classes in this code, otherwise it will be hard to use an #aop

src/Instrument/Transformer/CachingTransformer.php
lisachenko: I'll think a little about this (on my way to the home now and preparing for podcasting)

src/Instrument/ClassLoading/CachePathResolver.php
k-meister, line 18: Sometimes I think that there are too many of them :)
..Made some changes, hope now it's better

src/Instrument/Transformer/CachingTransformer.php
```diff
@@ -87,10 +96,11 @@ public function transform(StreamMetaData $metadata)
         }
 
         $originalUri = $metadata->uri;
-        $cacheUri    = str_replace($this->rootPath, $this->cachePath, $originalUri);
+        $cacheUri    = $this->container->get('aspect.cache.path.resolver')->getCachePathForResource($originalUri);
```
lisachenko: You should store this instance of service in the constructor. Personally, I'd like to see this implemented with DIC, but it's not possible for the moment. However, accessing local variable will be much faster than asking the container everytime.

src/Instrument/Transformer/FilterInjectorTransformer.php
```diff
@@ -97,7 +104,7 @@ public static function rewrite($originalResource, $originalDir = '')
             return self::PHP_FILTER_READ . self::$filterName . "/resource=" . $resource;
         }
 
-        $newResource = str_replace($appDir, $cacheDir, $resource);
+        $newResource = self::$kernel->getContainer()->get('aspect.cache.path.resolver')->getCachePathForResource($resource);
```
lisachenko: Expensive method call, please, do not use container as service locator every time. This method will be called thousands of time, so this can be too slow. To optimize this, just store cache resolver instance directly.

src/Instrument/Transformer/FilterInjectorTransformer.php
```diff
@@ -39,30 +40,36 @@ class FilterInjectorTransformer implements SourceTransformer
...
+     * @param AspectKernel $kernel Kernel to use for configuration
      * @param string $filterName Name of the filter to inject
      *
      * @throws \RuntimeException if filter was configured early
      */
-    protected static function configure(array $options, $filterName)
+    protected static function configure(AspectKernel $kernel, $filterName)
     {
-        if (self::$options) {
+        if (!is_null(self::$kernel)) {
```
lisachenko: Do not use `is_null()` check, PHP is dynamic language, so `if (!self::$kernel)` will be better for me.

k-meister:
Implemented resolver storing in transformers

lisachenko:
Thanks for this improvement! Merged into master (aka 1.0.0-dev).

k-meister:
Thank you for your job and inspiration :)

By the way, did you get an email from me, concerning some php-token-reflection issues?
There are certain inheritance/trait-mixin schemes that make reflector (and Go, respectively) throw an exception and break the code. It can be a serious limitation for applicability of your library.
Illustration of problem: https://github.com/Andrewsville/PHP-Token-Reflection/compare/develop...k-meister:fix-multiple-trait-import
Can we discuss it?

lisachenko:
Oh sorry, missed your email occasionally, just discovered it...Shame for me

Definitely I will go away from Php-Token-Reflection, because it isn't supported now and doesn't contain important fixes for upcoming PHP5.6 and PHP7. Thus, I am ready for discussion and implementation. 

I have a several thoughts for now:
- Make a fork for Php-Token-Reflection and support it as a component for framework
- Switch to [ApiGen/ElementReflection](https://github.com/ApiGen/ElementReflection) which is developed now by @TomasVotruba as fork of Php-Token-Reflection
- Directly use [nikic/PHP-Parser](https://github.com/nikic/PHP-Parser) as a reliable AST parser.

I like the first approach as fast (because framework almost don't care about PHP expressions inside method body) and the last approach as reliable, but more slow.

Moved this discussion to the #210

PR #258, opened 2016-02-28
Symfony integration changes for cache-warming

lisachenko:
This PR is needed for correct cache invalidation with Symfony2 framework


lisachenko:
This PR is preparation for fixing https://github.com/goaop/goaop-symfony-bundle/issues/3

PR #262, opened 2016-03-16
AopComposerLoader: prevent notices if no prefixes present

kornrunner:
We're using multiple composer-generated autoloader files.
One of them doesn't have any prefixes, so this patch prevents notices such as:

```
Notice: Undefined index: Dissect in vendor/goaop/framework/src/Instrument/ClassLoading/AopComposerLoader.php on line 76
Notice: Undefined index: Doctrine\Common\Annotations\ in vendor/goaop/framework/src/Instrument/ClassLoading/AopComposerLoader.php
```

### DIFF
--- a/src/Instrument/ClassLoading/AopComposerLoader.php
+++ b/src/Instrument/ClassLoading/AopComposerLoader.php
@@ -72,9 +72,11 @@ class AopComposerLoader
         $prefixes     = $original->getPrefixes();
         $excludePaths = $options['excludePaths'];
 
-        // Let's exclude core dependencies from that list
-        $excludePaths[] = $prefixes['Dissect'][0];
-        $excludePaths[] = substr($prefixes['Doctrine\\Common\\Annotations\\'][0], 0, -16);
+        if (!empty($prefixes)) {
+            // Let's exclude core dependencies from that list
+            $excludePaths[] = $prefixes['Dissect'][0];
+            $excludePaths[] = substr($prefixes['Doctrine\\Common\\Annotations\\'][0], 0, -16);
+        }
 
         $fileEnumerator       = new Enumerator($options['appDir'], $options['includePaths'], $excludePaths);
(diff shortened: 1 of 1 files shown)


lisachenko:
It's very strange issue, could you give me some more information about how it's possible to have a multiple composer loaders in parallel?

kornrunner:
Say you've got following structure:

```
app/
  one/
    vendor/
      autoload.php
    composer.json
    composer.lock
  two/
    vendor/
      autoload.php
    composer.json
    composer.lock
vendor/
  autoload.php
composer.json
composer.lock
```

There are multiple applications (`one` and `two`, etc) in given project directory. They use some global requirements specified in `composer.json` that resides in project root. However, every app has some specific requirements or just (ab)uses composer for generating autoloader.

Every frontend (or application) loads "shared" composer (root one) and it's own composer autooader instance, like so:

```
require_once 'vendor/autoload.php';
require_once 'app/one/vendor/autoload.php';

// application code continues
```

lisachenko:
Wow ) Never used such complex configuration, typically one application == one composer. Ok, but now I can think about your case, please wait a little bit.

One more question: do you use 1.x version or master (aka 2.x-dev) of `goaop/framework`?

kornrunner:
Issue is present with 1.x, haven't tried with master. It worked fine with 0.6.\* versions (actually, `goaop/framework` is a dependancy via https://github.com/Codeception/AspectMock in our project)

schlessera:
I have the same issue right now. Use case is a WordPress install where several plugins use composer dependencies. Although this will all be pulled in at the site level for production, it is handled at the individual plugin level during development, so there are multiple Composer roots.

schlessera:
I have provided an alternative pull request, as this one misses the case where `$prefixes` does exist, but does not contain the required dependencies.

lisachenko:
I'm prefer #263

PR #263, opened 2016-03-17
Only add existing core dependencies path exclusions in `AopComposerLoader`

schlessera:
Only add core dependencies path exclusions in `AopComposerLoader` when they can indeed been found within the current `ClassLoader`, to avoid PHP notices in scenarios where Composer provides multiple `ClassLoader`s.

Alternative fix to #262 , which missed the case where `$prefixes` exists, but does not contain the dependencies to be excluded.

### DIFF
--- a/src/Instrument/ClassLoading/AopComposerLoader.php
+++ b/src/Instrument/ClassLoading/AopComposerLoader.php
@@ -72,9 +72,15 @@ class AopComposerLoader
         $prefixes     = $original->getPrefixes();
         $excludePaths = $options['excludePaths'];
 
-        // Let's exclude core dependencies from that list
-        $excludePaths[] = $prefixes['Dissect'][0];
-        $excludePaths[] = substr($prefixes['Doctrine\\Common\\Annotations\\'][0], 0, -16);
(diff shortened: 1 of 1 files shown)


src/Instrument/ClassLoading/AopComposerLoader.php
```diff
@@ -72,9 +72,15 @@ public function __construct(ClassLoader $original, AspectContainer $container, a
         $prefixes     = $original->getPrefixes();
         $excludePaths = $options['excludePaths'];
 
-        // Let's exclude core dependencies from that list
-        $excludePaths[] = $prefixes['Dissect'][0];
-        $excludePaths[] = substr($prefixes['Doctrine\\Common\\Annotations\\'][0], 0, -16);
+        if (!empty($prefixes)) {
+            // Let's exclude core dependencies from that list
+            if (array_key_exists('Dissect', $prefixes)) {
```
lisachenko, line 77: Could you please use `isset` instead of `array_key_exists()`? And also target 1.x for merge instead of master.

schlessera:
Replaced by #264 .

PR #264, opened 2016-03-17
Only access existing prefix indexes when exluding paths.

schlessera:
Only add core dependencies path exclusions in `AopComposerLoader` when they can indeed been found within the current `ClassLoader`, to avoid PHP notices in scenarios where Composer provides multiple `ClassLoader`s.

Replaces #263, to target 1.x branch.


lisachenko:
Merged, thanks!

Issue #232, opened 2015-07-03
Indirect modification of properties causes no effect


Ocramius:
As per today's testing, I found that GO-AOP-PHP proxies are quite weak on magic property access.

Specifically, consider following class:

``` php
<?php

class Foo
{
    private $bar = [];
}

$foo = new Foo;
$foo->bar['baz'] = 'tab';
```

Defining a property access interceptor for this property will likely cause an output like:

```
Notice: Indirect modification of overloaded property Foo::$bar has no effect in /in/lLa8o on line 14 
```

This is because the generated code looks like:

``` php
<?php

class Foo
{
    private $bar = [];

    public function __get($name)
    {
        return $this->bar;
    }
}

$foo = new Foo;
$foo->bar['baz'] = 'tab';
```

The problem is that `__get` is not defined by reference.

I'm taking a ProxyManager generated proxy as a reference implementation (mainly because I tested a lot of aspects of it): https://gist.github.com/Ocramius/55cd7f709c91674add33

What does this mean for the API?
- `& $generated#__get`
- `& Go\Aop\Framework\ClassFieldAccess#__invoke()`
- `& Go\Aop\Framework\ClassFieldAccess#proceed()`

Ocramius:
In general, this needs quite extensive unit testing, specifically on the following operations:
- getting a property
- setting a property
- isset-checking a property 
- unsetting a property
- getting a property array key
- setting a property array key
- isset-checking a property array key
- unsetting a property array key

Ocramius:
Some tests to take as a reference implementation: https://github.com/Ocramius/ProxyManager/blob/7badedb5f8afaaccd496a096daadb19d7bf2b374/tests/ProxyManagerTest/Functional/LazyLoadingGhostFunctionalTest.php#L224-L285

lisachenko:
Thanks, @Ocramius, this is known issue #54, actually it isn't fully baked now like in your ProxyManager. I have never seen before the real use-cases for intercepting property access, because it's very rare case ) Your check is real first one that requires an interceptions of properties. 

My proxy is very limited indeed (no `__isset`, `__unset`, no reference passing), so a lot of thing to do and test.

So can I just your library to generate a proxy with my method body? This can be slower, but will more reliable and extendable.

Ocramius:
As it currently stands, the problem is that `ProxyManager` uses runtime reflection, which is not compatible with how GO-AOP-PHP works (runtime code generation).

[...]

PR #281, opened 2016-05-03
Rewrite property interceptors to be by reference

### LINKED ISSUE #54: Notice: Indirect modification of overloaded property ... has no effect
There is a class with protected property `$tasks` which is an array:

``` php
class TaskManager {

    protected $tasks = array();
}
```

There is also an advice which intercepts an access to this property:

``` php
    /**
     * Access interceptor
     * @param FieldAccess $property Joinpoint
     *
     * @Around("access(protected TaskManager->tasks)")
     * @return mixed
     */
    public function tasksChangeWatcher(FieldAccess $property) {
        static $tasksCount = 0;
        $value = $property->proceed();
        $tasksCnt = count($value);
        if ($tasksCount != $tasksCnt) {
            $tasksCount = $tasksCnt;
            Logger::getInstance()->addInfo("Number of tasks in the queue: $tasksCount");
        }
        return $value;
    }
```

During changing the value of this property in the application source code, php raises a warning:

```
Notice: Indirect modification of overloaded property TaskManager::$tasks has no effect in ...TaskManager.class.php on line 80
```

So if an advice applied to the private/protected property which is an array, then main source code cannot change it indirectly.

lisachenko:
This PR changes the property interception logic. Now all common logic is added by trait, additionally, all properties and methods for accessing properties now works "by reference", allowing for indirect changes and array key access.

This PR also includes a scope checker method to prevent an access to protected fields from external scope.
This should resolve #54 and #232

### DIFF
--- a/demos/Demo/Aspect/PropertyInterceptorAspect.php
+++ b/demos/Demo/Aspect/PropertyInterceptorAspect.php
@@ -32,10 +32,18 @@ class PropertyInterceptorAspect implements Aspect
      */
     public function aroundFieldAccess(FieldAccess $fieldAccess)
     {
-        $value = $fieldAccess->proceed();
-        echo "Calling Around Interceptor for ", $fieldAccess, ", value: ", json_encode($value), PHP_EOL;
+        $isRead = $fieldAccess->getAccessType() == FieldAccess::READ;
+        // proceed all internal advices
+        $fieldAccess->proceed();
 
-        // $value = 666; You can change the return value for read/write operations in advice!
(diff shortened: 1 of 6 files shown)


lisachenko:
Be aware, that property interceptors advices now work in another way. You can not use a result from `$jointPoint->proceed()` to access the current value of property or value to be set. Now `$joinPoint->getValue()` and `$joinPoint->getValueToSet()` should be used to access property values. In case you want to change the value, just return this value by reference.

PR #282, opened 2016-05-04
Property names documented with @property should be prefixed with a $ sign

natechicago:
I also clarified the documentation for the `source` element by suggesting that it contains the _contents_ of the stream.  But please let me know if this isn't always accurate and should be removed.

### DIFF
--- a/src/Instrument/Transformer/StreamMetaData.php
+++ b/src/Instrument/Transformer/StreamMetaData.php
@@ -17,18 +17,18 @@ use InvalidArgumentException;
 /**
  * Stream metadata object
  *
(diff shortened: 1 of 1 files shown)


natechicago:
I also clarified the documentation for the `source` element by suggesting that it contains the *contents* of the stream.  But please let me know if this isn't always accurate and should be removed.

lisachenko:
Merged, thanks! Thank you for contribution!

PS. I thinking more about the fact that this class should be a plain DTO with public properties and without ArrayAccess. Maybe, you can do this too as well? (Need to target 1.x branch)

PR #283, opened 2016-05-04
Fixing documentation in AopComposerLoader::findFile($class)

natechicago:
This class doesn't extend/implement another class, so using `{@inheritDoc}` on one of its methods isn't really appropriate.


lisachenko:
Merged, thanks!

PR #295, opened 2016-09-29
Use realpath() to resolve unresolved path from composer

RLasinski:
Since PR #5174 (https://github.com/composer/composer/pull/5174) the paths from composer can have dots in the paths. This isn't handled by AopComposerLoader.
CachePathManager tries to replace a resolved appPath with an unresolved resourcePath.
e.g. /app/ .... /vendor/composer/../../app

My solution isn't optimal, because _realpath()_ is called for every included file.

### DIFF
--- a/src/Instrument/ClassLoading/AopComposerLoader.php
+++ b/src/Instrument/ClassLoading/AopComposerLoader.php
@@ -155,6 +155,7 @@ class AopComposerLoader
         $file = $this->original->findFile($class);
 
         if ($file) {
+            $file = realpath($file)?:$file;
             $cacheState = isset($this->cacheState[$file]) ? $this->cacheState[$file] : null;
             if ($cacheState && $isProduction) {
                 $file = $cacheState['cacheUri'] ?: $file;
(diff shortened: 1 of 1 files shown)


lisachenko:
Hi, your path looks valid for PHP>=5.6 after changes in the composer. However, could you please replace `realpath()` with `PathResolver::realpath()` that can work even with non-standard paths from phar archivers, etc...

jcchavezs:
+1

src/Instrument/ClassLoading/AopComposerLoader.php
```diff
@@ -155,6 +155,7 @@ public function findFile($class)
         $file = $this->original->findFile($class);
 
         if ($file) {
+            $file = realpath($file)?:$file;
```
lisachenko: Could you please replace `realpath()` with `PathResolver::realpath()`?

lisachenko (changes requested):

lisachenko (approved):

lisachenko:
Merged, thanks!

PR #296, opened 2016-10-24
stay compatible with PHP 5.5

infabo:
Doctrine just launched Annotations 1.3.0 which is now PHP 5.6+.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     "require": {
         "php": ">=5.5.0",
         "jakubledl/dissect": "~1.0",
-        "doctrine/annotations": "~1.0",
+        "doctrine/annotations": "<1.3.0",
         "andrewsville/php-token-reflection": "~1.4"
     },
 


lisachenko:
Hi! Why do you need to put a restriction into the framework? Maybe it will be better just to put this requirement into your application?

```
composer require doctrine/annotations:<1.3.0
```

Then composer will solve all requirements automatically

infabo:
I already put this restriction into my application. Just wanted to be a nice guy. Goaop framework 1.x branch is NOT compatible with PHP 5.6+, so WHY require a library which needs PHP 5.6 - sry to say, but this is silly. I'm out.

lisachenko:
Sorry, if my answer was not so polite, just wanted to ask why do you want to put such restriction. I agree, that version 1.x is not maintained and can not be used for 5.6 with all new features due to the limitations of `andrewsville/php-token-reflection` packet. 

If you propose me to restrict php version in this PR to <5.6, then it can be merged without any additional questions. Even I don't like the idea to put any additional restrictions for 1.x branch because it's not maintained and used for legacy project to keep support.

Hope, you are ok, because I'm really appreciate each PR to the framework and always try to communicate with authors for discussions about changes in the PR before merging them into the project.

PR #300, opened 2016-11-22
Add path wildcard feature to Enumerator

icanhazstring:
Add wildcard possibility for `includePaths` and `excludePaths`

### DIFF
--- a/src/Instrument/FileSystem/Enumerator.php
+++ b/src/Instrument/FileSystem/Enumerator.php
@@ -82,17 +82,14 @@ class Enumerator
         $includePaths = $this->includePaths;
         $excludePaths = $this->excludePaths;
 
-        // Parse wildcard paths
-        $excludePaths = array_map([$this, 'parseWildcardPath'], $excludePaths);
-        $includePaths = array_map([$this, 'parseWildcardPath'], $includePaths);
-
         return function (\SplFileInfo $file) use ($rootDirectory, $includePaths, $excludePaths) {
+
             if ($file->getExtension() !== 'php') {
                 return false;
             };
 
-            $fullPath = $file->getRealPath();
-            $filePath = $file->getPath();
+            $fullPath = $this->getFileFullPath($file);
+            $dir      = dirname($fullPath);
             // Do not touch files that not under rootDirectory
             if (strpos($fullPath, $rootDirectory) !== 0) {
                 return false;
@@ -101,7 +98,7 @@ class Enumerator
             if (!empty($includePaths)) {
                 $found = false;
                 foreach ($includePaths as $includePattern) {
-                    if (preg_match('%' . $includePattern . '%', $filePath)) {
+                    if (fnmatch($includePattern, $dir)) {
                         $found = true;
                         break;
                     }
@@ -112,7 +109,7 @@ class Enumerator
             }
 
             foreach ($excludePaths as $excludePattern) {
-                if (preg_match('%' . $excludePattern . '%', $filePath)) {
+                if (fnmatch($excludePattern, $dir)) {
                     return false;
                 }
             }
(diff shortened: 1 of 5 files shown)


lisachenko:
Hi! Thank you for your job! I'll review it soon

src/Instrument/FileSystem/Enumerator.php
```diff
@@ -78,24 +79,29 @@ public function enumerate()
...
+            if (strpos($fullPath, $rootDirectory) !== 0) {
                 return false;
             }
 
             if (!empty($includePaths)) {
                 $found = false;
-                foreach ($includePaths as $includePath) {
-                    if (strpos($realPath, $includePath) === 0) {
+                foreach ($includePaths as $includePattern) {
+                    if (preg_match('%' . $includePattern . '%', $filePath)) {
```
lisachenko: I have a worry, that `preg_match()` could be slow for checking a lot of files during class loading of big framework. Could you please check, if `fnmatch()` could be faster than `preg_match()`.

src/Instrument/FileSystem/Enumerator.php
```diff
@@ -105,13 +111,31 @@ public function getFilter()
...
     }
+
+    /**
+     * Replace possible wildcard with proper regex patterns
+     *
+     * @param string $path
+     *
+     * @return string
+     */
+    protected function parseWildcardPath($path)
```
lisachenko: This function can be replaced with built-in `fnmatch()`, see [fnmatch() function](http://php.net/manual/en/function.fnmatch.php)

tests/Go/Instrument/FileSystem/EnumeratorTest.php
```diff
@@ -0,0 +1,135 @@
...
+     * @throws \Exception
+     * @return void
+     */
+    public static function setUpBeforeClass()
+    {
+        if (!defined('TEST_DIRECTORY')) {
+            throw new \Exception('TEST_DIRECTORY not set, check your phpunit.xml');
+        }
+
+        static::$fixtureBasePath =  realpath(TEST_DIRECTORY) . DIRECTORY_SEPARATOR . 'fixtures';
```
lisachenko: Oh, it's an armed gun ) Because for several cases `realpath()` can return false which will equal to the empty string. Maybe we can use VFS filesystem for safe tests?

tests/Go/Instrument/FileSystem/EnumeratorTest.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Clean fixture paths after tests are done
+     *
+     * @return void
+     */
+    public static function tearDownAfterClass()
+    {
+        exec('rm -rf ' . static::$fixtureBasePath);
```
lisachenko: Very dangerous command, I think VFS will be better to use, because who knows if `static::$fixtureBasePath` will equal to the '/'?

icanhazstring:
@lisachenko can you merge this into the 1.x version line as well?
Problem ist for example we currently can only use PHP<=5.5 - but with the 2.x version line PHP>5.6 is required and the PR could be easily merged in both (no conflicts)

So this needs to be merged into both versions - if possible

lisachenko (changes requested):

lisachenko:
@icanhazstring changed the base to 1.x, please don't forget to rebase your local branch as well

src/Instrument/FileSystem/Enumerator.php
icanhazstring: Problem is that fnmatch does not work with multiple subfolders. so `/base/*/test` only includes `/base/sub/test` but not `/base/sub/sub/test`

lisachenko: Please, see: https://3v4l.org/iPR39#output Your case is second line

icanhazstring: Well... ok. I'll rewrite it then :)

icanhazstring:
@lisachenko reworked the tests using vfs and using `fnmatch()` for testing path against pattern

lisachenko (approved):

lisachenko:
Squashed and merged! Thank you very much! :+1:

PR #301, opened 2016-11-28
Add phpunit as dependency

icanhazstring:
This will add phpunit as a development dependency so the user don't have to install phpunit system wide. This is especially useful if the unittests depend on a certain phpunit version but the user can't update the systemwide installation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
   - composer install
 
 script:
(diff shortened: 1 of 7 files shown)


icanhazstring:
Will do a request for 2.x as well. I currently can't require phpunit:^5.0 due to php 5.5 limitation on my current system.

lisachenko:
Squashed and merged! Thank you!

PR #305, opened 2016-12-12
Propose PHP <= 5.5

infabo:
Propose PHP <= 5.5 for 1.x branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "license": "MIT",
 
     "require": {
-        "php": ">=5.5.0",
+        "php": "<=5.5.99",
         "jakubledl/dissect": "~1.0",
         "doctrine/annotations": "<1.3.0",
         "andrewsville/php-token-reflection": "~1.4"
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     "require": {
         "php": ">=5.5.0",
         "jakubledl/dissect": "~1.0",
-        "doctrine/annotations": "~1.0",
+        "doctrine/annotations": "<1.3.0",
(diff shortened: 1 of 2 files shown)


lisachenko:
Hi, there are errors in the composer.lock, please resolve them.

composer.json
```diff
@@ -8,9 +8,9 @@
     "license": "MIT",
 
     "require": {
-        "php": ">=5.5.0",
+        "php": "<=5.5.99",
```
lisachenko: It's better to use ~5.5.0 here.

composer.json
```diff
@@ -8,9 +8,9 @@
...
         "jakubledl/dissect": "~1.0",
-        "doctrine/annotations": "~1.0",
+        "doctrine/annotations": "<1.3.0",
```
lisachenko: Please, switch to ~1.2.0

lisachenko (changes requested):

composer.json
infabo: This would restrict the 1.x branch to PHP 5.5 only. Any reasons why not support PHP 5.3 / 5.4?

composer.json
infabo: acknowledged.

infabo:
Can't find any issues in composer.lock.

composer.json
lisachenko: 5.3 and 5.4 are not supported, anyone who wants to use an old version, can switch to 0.x versions.

lisachenko:
Merged, thanks

lisachenko:
Also, removed the `composer.lock` file from the repository.

PR #309, opened 2017-01-05
Add support for nullable types for the PHP7.1

lisachenko:


src/Proxy/AbstractProxy.php
```diff
@@ -204,11 +205,13 @@ protected function prepareArgsLine(ReflectionFunctionAbstract $functionLike)
...
-        $reflectionReturnType = PHP_VERSION_ID >= 50700 ? $functionLike->getReturnType() : '';
+        $reflectionReturnType = PHP_VERSION_ID >= 70000 ? $functionLike->getReturnType() : '';
         $modifiersLine        = '';
         if ($reflectionReturnType) {
-            $nsPrefix             = $reflectionReturnType->isBuiltin() ? '' : '\\';
-            $reflectionReturnType = $nsPrefix . (string)$reflectionReturnType;
+            $nullablePrefix = $reflectionReturnType->allowsNull() ? '?' : '';
+            $nsPrefix       = $reflectionReturnType->isBuiltin() ? '' : '\\';
+
+            $reflectionReturnType = $nullablePrefix . $nsPrefix . (string) $reflectionReturnType;
```
smuggli, line 214: @lisachenko Are you sure that this works as desired? I havn't tested the code but in case I am using PHP 7.0.x, have a method with a return type and a signature like "($parameter = null)", a nullable operator will be added. This would break the code because nullable were added in PHP 7.1 like mentioned in the title. Missed I something?

lisachenko: @smuggli nice catch! In this part of PR everything is ok, because return type can be nullable only in 7.1, for 7.0 return types `allowsNull()` will be false for every case.

But there is a bug in the first part with checking for nullable parameters. So, yes, confirmed as a bug for 7.0.

smuggli: Ahh okay, thanks for the quick response.

PR #326, opened 2017-04-24
fnmatch bugfix when used on Windows

bhoehl:
@lisachenko 
Windows has pathes like c:\projects\....
If a path contains backslashes fnmatch will fail.


lisachenko:
Thanks! Merged into 1.x and master.

Closed via https://github.com/goaop/framework/commit/346cec9e29d8c3c0103167c4d32ad2618c198080

bhoehl:
@lisachenko 
when will there be a new release containing this ?
I ask as we are using AspectMock and we do have to patch those files all the time

lisachenko:
Prepared patch tags: 2.1.1 and 1.2.1

Issue #327, opened 2017-07-07
[Known-issue] Doctrine entities could not be woven 


TheCelavi:
Doctrine entities can not be weaved - class metadata gets all messed up.

Engine generates `Entity__AopProxied`, which is copy of original class, and then generates class `Entity` that extends `Entity__AopProxied` and contains weaved methods/properties.

Issue is that `Entity__AopProxied` contains doctrine metadata, so Doctrine actually generates such metadata that `Entity__AopProxied` is main class and every relation is bound to that class, not `Entity`.

Side effects are that, per example, table name gets suffix "__aop_proixed", while relations gets all screwed up, so any join is broken because different aliases are generated with SQL walker. Lazy loading does not work as well.

Possible solution would be (in order to preserve BC compatibility) to provide possibility to specify different method of weaving for certain classes. Per example, in this case, in order to work as it should, `Entity__AopProxied` should not be generated and inherited, it should be generated only `Entity` class.

Yes, maybe we will loose nice debugging and breakpoints, however, for now, we can sacrifice that for working code.

TheCelavi:
Note - I am using annotations to map entities.

TheCelavi:
Tried to use different mapping method - just for testing purposes, no love. Which is reasonable, Entity__AopProxied is not valid superclass.... 

Here is the example of wrong SQL: 

An exception occurred while executing 'SELECT t1.id AS id_2, t1.feed_item_id AS feed_item_id_3, t1.url AS url_4, t1.platform AS platform_5, t1.views AS views_6, t1.likes AS likes_7, t1.dislikes AS dislikes_8, t1.comments AS comments_9, t1.shares AS shares_10, t1.created_at AS created_at_11, t1.updated_at AS updated_at_12, t1.created_by AS created_by_13, t1.updated_by AS updated_by_14, t1.offer_id AS offer_id_15 FROM report_item t1 WHERE t0.offer_id = ?' 

SQLSTATE[42S22]: Column not found: 1054 Unknown column 't0.offer_id' in 'where clause'").

Note how generated alias for report_item is t1, but for where statement is t0 ? This is simple, lazy loaded relation.

TheCelavi:
Ok, this does not allow me to sleep at all. By playing around, going trough code, I think that there is a workaround in for this situation, however, Doctrine mapping must be modified (tried with hardcoding just to test theory, works).

[...]

PR #328, opened 2017-07-09
Added workaround for #327

TheCelavi:
As discussed, here is the pull request with Doctrine entities weaving workaround. (see [https://github.com/goaop/framework/issues/327](https://github.com/goaop/framework/issues/327))

- Added doctrine listener that will correct generated metadata
- Added tests (sorry, no 100% code coverage, method `getTraits` is private, so it would be very hard to provide it. In original library, code has 100% code coverage, but I didn't wanted to add more dependencies)
- Modified README.md, added section about this issue and support

TODO:
- Add support for Symfony' bundle: Event subscriber will be always disabled by default (in order to improve performance). When it is enabled, compiler pass will check if there is a Doctrine ORM enabled in Symfony project, and if there is, it will register event subscriber.

### DIFF
--- a/src/Bridge/Doctrine/MetadataLoadInterceptor.php
+++ b/src/Bridge/Doctrine/MetadataLoadInterceptor.php
@@ -1,0 +1,153 @@
+<?php
+/*
+ * Go! AOP framework
+ *
+ * @copyright Copyright 2012, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+namespace Go\Bridge\Doctrine;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
+use Doctrine\ORM\Events;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Go\Core\AspectContainer;
+
+/**
+ * Class MetadataLoadInterceptor
+ *
+ * Support for weaving Doctrine entities.
+ *
+ * @package Go\Bridge\Doctrine
+ */
+final class MetadataLoadInterceptor implements EventSubscriber
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSubscribedEvents()
+    {
+        return [
+            Events::loadClassMetadata
+        ];
+    }
+
+    /**
+     * Handles \Doctrine\ORM\Events::loadClassMetadata event by modifying metadata of Go! AOP proxied classes.
+     *
+     * This method intercepts loaded metadata of Doctrine's entities which are weaved by Go! AOP,
+     * and denotes them as mapped superclass. If weaved entities uses mappings from traits
+     * (such as Timestampable, Blameable, etc... from https://github.com/Atlantic18/DoctrineExtensions),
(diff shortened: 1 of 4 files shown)


src/Bridge/Doctrine/MetadataLoadInterceptor.php
```diff
@@ -0,0 +1,153 @@
...
+use Doctrine\ORM\Events;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Go\Core\AspectContainer;
+
+/**
+ * Class MetadataLoadInterceptor
+ *
+ * Support for weaving Doctrine entities.
+ *
+ * @package Go\Bridge\Doctrine
```
lisachenko: Please, drop this auto-generated tag.

tests/Go/Aop/Bridge/Doctrine/MetadataLoadInterceptorTest.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+trait SimpleTrait
+{
+    private $mappedField;
+}
+
+
+class Entity__AopProxied
```
TheCelavi: I can not fix this, except to pull out to separate files - however, this class is used only for this test suite, so usual practice is to place it inline within test suite. 

Class name must stay same, that is clear.

tests/Go/Aop/Bridge/Doctrine/MetadataLoadInterceptorTest.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->assertEquals(0, count($metadata->table));
+        $this->assertNull($metadata->customRepositoryClassName);
+
+        $this->assertFalse(isset($metadata->fieldMappings['mappedField']));
+        $this->assertFalse(isset($metadata->fieldNames['mapped_field']));
+        $this->assertFalse(isset($metadata->columnNames['mappedField']));
+    }
+}
+
+trait SimpleTrait
```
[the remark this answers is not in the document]
TheCelavi: I can not fix this, except to pull out to separate files - however, this class is used only for this test suite, so usual practice is to place it inline within test suite.

src/Bridge/Doctrine/MetadataLoadInterceptor.php
```diff
@@ -0,0 +1,153 @@
...
+         * @var ClassMetadata $metadata
+         */
+        $metadata = $args->getClassMetadata();
+
+        if (1 === preg_match(sprintf('/.+(%s)$/', AspectContainer::AOP_PROXIED_SUFFIX), $metadata->name)) {
+
+            $metadata->isMappedSuperclass = true;
+            $metadata->isEmbeddedClass = false;
+            $metadata->table = [];
+            $metadata->customRepositoryClassName = null;
```
lisachenko: it would be nice to have variables aligned.

lisachenko (changes requested):

tests/Go/Aop/Bridge/Doctrine/MetadataLoadInterceptorTest.php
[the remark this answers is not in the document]
lisachenko: We can ignore this, ok

lisachenko (approved):

lisachenko:
Squashed and merged, thanks!

PR #331, opened 2017-07-10
Fixed #330 - commands are modified, they can be owerwritten now

### LINKED ISSUE #330: Commands should be modified in order to support other frameworks
Commands:

- `debug:advisor`
- `debug:aspect`
- `cache:warmup:aop`

are extremely useful, but they are not exposed, per example, in Go Aop Symfony bundle. 

This ought to be corrected, and it can be done with ease. Class `Go\Console\Command\BaseAspectCommand` should be modified, `execute()` method body should go to some protected method in order to allow modification of it. That method is only in charge to get Apsect Kernel, and in Symfony, per example, Aspect Kernel can be acquired differently. 

After that, it is required for each framework to bridge usage of those commands.

This is easy pick, I can fix core, and provide Symfony bridge.

TheCelavi:
This PR will enable us to expose commands for other frameworks (especially in Symfony).

These commands are very useful and should be available to all bridges/adapters out-of-the-box.

TODO:
    - Expose commands in Symfony AOP bundle (I can do that)
    - Expose commands in other FWs

### DIFF
--- a/src/Console/Command/BaseAspectCommand.php
+++ b/src/Console/Command/BaseAspectCommand.php
@@ -37,16 +37,26 @@ class BaseAspectCommand extends Command
     }
 
     /**
-     * {@inheritDoc}
+     * Loads aspect kernel.
+     *
+     * Aspect kernel is loaded by executing loader and fetching singleton instance.
+     * If your application environment initializes aspect kernel differently, you may
+     * modify this metod to get aspect kernel suitable to your needs.
+     *
+     * @param InputInterface $input
+     * @param OutputInterface $output
      */
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function loadAspectKernel(InputInterface $input, OutputInterface $output)
     {
         $loader = $input->getArgument('loader');
         $path   = stream_resolve_include_path($loader);
         if (!is_readable($path)) {
             throw new \InvalidArgumentException("Invalid loader path: {$loader}");
         }
+
+        ob_start();
         include_once $path;
+        ob_clean();
 
         if (!class_exists(AspectKernel::class, false)) {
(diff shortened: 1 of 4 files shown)


src/Console/Command/BaseAspectCommand.php
```diff
@@ -37,16 +37,26 @@ protected function configure()
...
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function loadAspectKernel(InputInterface $input, OutputInterface $output)
     {
         $loader = $input->getArgument('loader');
         $path   = stream_resolve_include_path($loader);
         if (!is_readable($path)) {
             throw new \InvalidArgumentException("Invalid loader path: {$loader}");
         }
+
+        ob_start();
```
TheCelavi, line 57: When $path is included, it's being executed. Usually, that is some front controller of web application and homepage is rendered, filling console with bunch of HTML code.

Surrounding it with  ob_start() and ob_clean() will just keep console nice and clean, making it more user friendly.

lisachenko: Yes, It would be nice just to drop this HTML stuff, agree with you.

lisachenko:
In general, commands should be registered in bridges. I'm not so happy about their implementation, but they can be very helpful to debug AOP or check what joinpoints are advised.

I have another one idea, just move all AOP kernel loading stuff to the composer. And define special composer hook that will write initialization file into the /vendor/goaop/bootstrap.php file during installation of dependencies. Commands can be used as-is without explicit inclusion of aspect kernel. But I can leave this for future versions, maybe your research and usage of framework will inspire me to do that :)

For now I will merge this PR. Thanks!

Issue #335, opened 2017-07-31
[BUG] Method matching positive on abstract classes which does not implements pointcuted method from interface


TheCelavi:
Consider usecase (precondition):

- Class/Interface structure: Interface I -> AbstractClass A implements I -> ConcreteClass C extends  A
- Interface defines method foo()
- Abstract class DOES NOT implements foo()
- ConcreteClass implements foo()

Consider that ALL stated Interface/Classes are within same namespace.

If matcher is execution(public *->foo()), abstract class gets weaved -> it is interpreted as if there is a method foo, but there is not.

Possible fix: check PHP Tokenizer and class metadata for better introspection of implemented methods.

lisachenko:
Hi! I can see for the abstract class, that it doesn't contain an abstract method after proxy generation. because `ClassProxy`'s method `overrideMethod` has special temporary check at line 382:

```php
        // temporary disable override of final methods
        if (!$method->isFinal() && !$method->isAbstract()) {
            $this->override($method->name, $this->getJoinpointInvocationBody($method));
        }
```

This check prevents interception of final or abstract methods (including methods from interfaces).

But I can agree that pointcut currently matches abstract and final methods too. Not sure if it's good or bad, just possible to match abstract method. It can be used for example for introduction pointcuts to introduce new properties, methods or body for interfaces.

TheCelavi:
> Not sure if it's good or bad, just possible to match abstract method.

I think it is bad, it should not match what can not be weaved. 

> It can be used for example for introduction pointcuts to introduce new properties, methods or body for interfaces.

Isn't that different story here? Introductions would introduce new Interfaces to a class, and via traits interfaces are implemented, right?

lisachenko:
> Isn't that different story here? Introductions would introduce new Interfaces to a class, and via traits interfaces are implemented, right?

Yes, for example, TYPO3 just adds properties and methods into specific class. But I've decided to switch to traits and interfaces to have an ability to put breakpoints in traits. And it also gives me control over line numbering in the classes which was important.

 > I think it is bad, it should not match what can not be weaved.

Ok, let's fix this 👍

lisachenko:
I haven't checked, but I think it will be possible to do this on grammar level by giving to the `ModifierMatcherFilter` negative filter for abstract modifier like that:

```php
$modifierMatcherFilter = new ModifierMatcherFilter();
$modifierMatcherFilter->notMatch(\ReflectionMethod::IS_ABSTRACT);
```

[...]

Issue #274, opened 2016-03-30
[Feature] Allow to match elements only in declared class


lisachenko:
Pointcut matching engine now hadles hierarchy elements correctly, see #150, but sometimes this leads to interesting results that aren't expected, for example matching method for several times in all children classes, for example #272 

There should be a pointcut syntax to restrict matching of elements only to the class where element was declared. Unfortunately, it won't be easy to change the signature for matching elements, but maybe some kind of context maybe injected into the point filter to allow more complex types of element matching.

lisachenko:
Closed in https://github.com/goaop/framework/commit/0dc33f990c35b6ea2914f508faf9de208626c335

Issue #272, opened 2016-03-30
Annotation advice is executed multiple times


schlessera:
### Problem

I have an logging aspect that attaches its advice through an annotation pointcut. I want to use this for general-purpose logging, and have built it so that I can configure the logging message with annotation properties.

However, when I use this on an object hierarchy, the aspect gets executed multiple times, once for each level of inheritance.

Is this a bug, a limitation, or am I doing something wrong?
### Example

Consider the following hierarchy:

```
User
   Better_User extends User
      Best_User extends Better_User
```

When I have a `Best_User` object, and call the `User::get_data()` method, the advice is executed 3x, once for each level of inheritance.

If I call `Better_User::other_method()`, it gets executed 2x, and so on...
### Current Code

``` PHP
/**
 * Advice around the pointcut.
 *
 * @since 1.0.0
 *
 * @Around("@execution(DT\Membership\Annotation\Loggable)")
 *
 * @param MethodInvocation $invocation The method that was invoked.
 * @throws Exception To forward the exception that was caught.
 * @return mixed
 */
public function aroundLoggable( MethodInvocation $invocation ) {

    $method = $invocation->getMethod()->name;

    $object   = $invocation->getThis();
    $loggable = $invocation->getMethod()->getAnnotation( Loggable::class );
    $message  = $loggable->getString( $object );
    $level    = $loggable->getLevel();

    // Enter method.
    $this->logger->$level(
        $message,
        (array) $invocation->getArguments()
    );

    try {
        // Execute the invoked method.
        $result = $invocation->proceed();

        // Success.
        $this->logger->$level(
            sprintf( __(
                '%1$s - SUCCESS.',
                'dt-membership' ),
                $message
            ),
            (array) $result
        );
    } catch ( Exception $e ) {

        // Failure.
        $this->logger->$level(
            sprintf( __(
                '%1$s - ERROR: %2$s.',
                'dt-membership' ),
                $message,
                $e->getMessage()
            )
        );
        throw $e;
    }

    return $result;
}
```

lisachenko:
Hi, I know what happens here. Method with annotation is present in the parent class, but it also matched with children classes as well, so method is intercepted 3 times in all hierarchy. You need somehow to restrict matching of this method in children classes. 

This feature was requested previously and was implemented in the engine level to intercept all parent methods and properties as well.

[...]

Issue #131, opened 2014-02-12
ISSUE, Calling a exteded function from parent class, doesn't work.


eriknyk:
Hi, If I have a class

```
class Greeting
{
    public hello($name)
    {
        echo "hello $name";
    }
}
```

And I extend this class

```
class Example extends Greeting
{
}
```

And I have defined on my aspect class:  

```
@Before("execution(public Example->*(*))")
```

And I execute a Example instance.

```
$e = new Example();
$e->hello("erik");
```

I doesn't work, the aspect class is not matching the class Example->*.

The unique way to do this works, was adding a method calling the parent class in Example class, so:

```
class Example extends Greeting
{
    public hello($name)
    {
        parent::hello($name);
    }
}
```

But it really doesn't a solution, because the idea of extending classes is do not repeat code.
Can you tell if it fixable?

Best Regards.

lisachenko:
Hi, Erik!

Yeap, this issue is one of the known limitation of framework. During the class loading process I don't know information about parent class, because it's not loaded into memory, so I can not override parent method in the child with custom logic. AOP intercepts only concrete methods in the concrete class and can not check methods in extending classes. 

However, it will be possible in the future to recursively load classes and interfaces before PHP to build an hierarchy, but I don't want to implement this logic right now, because it can break a lot of things.

eriknyk:
It would be more usable your framework, because  extensive projects have a lot of extending classes, that is my case,.. I can't use go-aop-php without this capability,..
I think that getting all extending methods, functions and properties recursively should works..
Well, I've waiting for news,.. 
Thanks a lot for your response.

Best Regards.

lisachenko:
@eriknyk you could try to use `feature/static-inheritance` branch, it should handle inheritance now )

lisachenko:
Resolved by #150

PR #337, opened 2017-08-08
Exclude abstract methods from pointcut matching #335

TheCelavi:
This is a first, working idea. I guess there is  a better approach. There are no tests written yet, however, I have tested internally and this do fixes my problem that I have described. 

Solution seams un-natural and hackish. I will explain in code general idea.

@lisachenko I would like from you to review it and propose a better/more elegant solution for this fix implementation.

Thanks.


src/Aop/Pointcut/ClassDeclaresPointcut.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Go\Aop\Pointcut;
+
+use Go\Aop\Pointcut;
+use Go\Aop\PointFilter;
+
+/**
+ * Pointcut that filters out all class method and properties which are not declared within class
+ */
+class ClassDeclaresPointcut implements Pointcut
```
TheCelavi: Idea is to have a pointcut matcher that will filter out all non-declared members and properties that are matched.

src/Aop/Pointcut/ClassDeclaresPointcut.php
```diff
@@ -0,0 +1,46 @@
...
+
+        return $point->getDeclaringClass()->getName() === $context->getName();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getKind()
+    {
+        return PointFilter::KIND_METHOD | PointFilter::KIND_PROPERTY;
```
TheCelavi: As I have understood - this will ensure that this pointcut matcher will be applied ONLY if subject of matching is class method and class property, right?

Otherwise, `matches` function will not be invoked, right?

src/Aop/Pointcut/PointcutParser.php
```diff
@@ -25,4 +26,12 @@ public function __construct(PointcutGrammar $grammar)
         $parseTable = include 'PointcutParseTable.php';
         parent::__construct($grammar, $parseTable);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function parse(TokenStream $stream)
+    {
+        return new AndPointcut(parent::parse($stream), new ClassDeclaresPointcut());
```
TheCelavi: Ok, so this is why I think that this solution is kinda hackish...

Whatever our parser gets as a pointcut filter, it has to be additionally filtered. So it has to be applied last.

`parent::parse($stream)` -> will provide a pointcut that will match something, and that something can be a collection of properties and methods, right?

So, whatever methods and properties we got, we have to apply `ClassDeclaresPointcut` matcher to filter out non-declared properties and methods.

I guess that Grammar should give us this expression, is that even possible?

TheCelavi: Oh, crap, I forgot that we can add a node visitor that would modify AST and provide us with same result: https://github.com/jakubledl/dissect/blob/master/docs/ast.md#traversing-the-ast

I can write node visitor to modify AST in order to move this logic from PointcutParser to node visitor and grammar. 

However - still need from you the verification of idea so I can proceed (write node visitor, tests, etc...)

TheCelavi: Ok, seams wrong -> I should only wrap original Pointcut if `getKind()` is PROPERTY or METHOD, right?

lisachenko:
Hi, Nikola! Thank you for researching this! But correct me if I'm wrong: you want to exclude all parent methods that aren't declared in current class, right?

If it's what you want to achieve, then there is special syntax for that: `&& !matchInherited()`. You can see this syntax in `PointcutParser` test case

```php
            // This will match all methods, but not inherited
            ['execution(public **->*(*)) && !matchInherited()'],
```

src/Aop/Pointcut/ClassDeclaresPointcut.php
lisachenko: Looks like it is already implemented in `MatchInheritedPointcut` 😄

src/Aop/Pointcut/ClassDeclaresPointcut.php
lisachenko: Yes, you are right, point filter effectively reduce the number of checks for concrete point filter. It's not ideal, but it's very efficient. `KIND_METHOD` filters only methods, `KIND_PROPERTY` filters properties in the class.

For future versions of framework I want to put this logic into `matches()` method. To be in one single place.

TheCelavi:
Yes. Ok, I got confused here for sure - why `!matchInherited()` is not applied by default?

Why would weaving of method/property occur on a class that does not declares it?

src/Aop/Pointcut/PointcutParser.php
lisachenko: You shouldn't touch the `PointcutParser` class at all, it's final :) Instead, all changes should be implemented in grammar and AST transforming, see `PointcutGrammar` class for examples how AST will be converted to the pointcut.

src/Aop/Pointcut/ClassDeclaresPointcut.php
TheCelavi: Yes, so it seams....

TheCelavi:
Ok, this is the issue of understanding what is the bug and what is not. This proposal is related to https://github.com/goaop/framework/issues/335. 

Lack of declaration existence check leads to issue #335

For me - there is no logic to weave method/property that is not declared within the class. If there is - I believe that it should be explicitly requested by pointcut expression, ie. `!matchInherited()` should be by default. 

This was my usecase:

    Interface X {
        public function foo();
    }

    abstract class Y implements X {

    }

    class Z extends Y {

        public function foo() {
        }
    }

Booth Z and Y classes are weaved. It should be only Z. And that should be by default, right?

lisachenko:
Previously developers asked me why there is no matching for parent methods, issue about matching for parent methods was described in #131 and later I was implemented logic to match parent properties and methods, see #272 and #274. New logic was introduced in PR #279.

This was logical for me to match methods from parent classes during weaving. And filtration should be performed via pointcut definition.

Regarding your question about switching matching to declared members only be default - it's discussable. Maybe we can introduce new option for the kernel to allow/disable matching with parent members for classes?

TheCelavi:
Ok, THANK YOU - now we are getting somewhere. In conclusion, we can not change this behaviour, it would result with BC break.

Conclusion is: weaving will occur always on all matching methods/properties, regardless if it is defined in class or some of the parent class (unless `! matchInherited()` is specified). **This behaviour must be documented!** So, this is on me.

Ok, so we have a bug, that is for sure, which is easy to define now:

1. Pointcut MUST NOT match abstract method of class
2. Pointcut MUST NOT match method of class that is declared in interface

Simple - what is not implemented in class hierarchy, can not be weaved.

Agree?

(if we can identify bug and agree upon what is a bug - we can solve it)

lisachenko:
> Pointcut MUST NOT match abstract method of class

Agree

> Pointcut MUST NOT match method of class that is declared in interface

Actually, it's the same as the first check, because method from an interface will be abstract too in reflection. So resulting statement will be: "Pointcut MUST NOT match abstract methods"

src/Aop/Support/ModifierMatcherFilter.php
```diff
@@ -47,6 +47,7 @@ class ModifierMatcherFilter implements PointFilter
     public function __construct($initialMask = 0)
     {
         $this->andMask = $initialMask;
+        $this->notMask = \ReflectionMethod::IS_ABSTRACT;
```
TheCelavi: @lisachenko Ok, it seams that you are right, modifier filter should have default value of \ReflectionMethod::IS_ABSTRACT for `notMask` .

Test passes, but what is worries me is that there are no a lot of integration/functional tests where we can test what we actually get (like what I did for console commands). 

What do you think about writing some functional tests that would have real-life fixtures for verification of weaving?

lisachenko: I feel myself shame for tests :) I'd vote for unit tests for each class and only several functional tests for checking that whole system is working as expected and still able to transform files and perform weaving. 

For this concrete class we can add just simple test case that modifier matcher filter by default doesn't match abstract methods.

One more caution: modifier filter is also used for property matching. Of course, "not abstract" for properties will be always true. But if we will use ModiferFilter for properties with negation, then it can be very unpredictable :)

src/Core/AdviceMatcher.php
```diff
@@ -166,6 +166,11 @@ private function getAdvicesFromAdvisor(
 
             $mask = ReflectionMethod::IS_PUBLIC | ReflectionMethod::IS_PROTECTED;
             foreach ($class->getMethods($mask) as $method) {
+                // abstract methods can not be weaved
```
TheCelavi, line 169: > One more caution: modifier filter is also used for property matching. Of course, "not abstract" for properties will be always true. But if we will use ModiferFilter for properties with negation, then it can be very unpredictable :)

You are soooo right! Ok, how about this? Simply - advice matcher will skip abstract methods for weaving, regardless of pointcut expression used?

lisachenko: This solution is perfect! 👍

lisachenko (approved):
Looking good now

tests/Go/Functional/BaseFunctionalTest.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Go\Functional;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Process\Process;
+
+abstract class BaseFunctionalTest extends TestCase
```
TheCelavi, line 9: @lisachenko  When you find a time, please review tests as well.

I have cleaned up code a little bit, setup a base for functional tests, write a required test as well as modified (cleaned up) command tests...

I have one more thing on my mind, however, even with this setup it is going to be quite easy to write a bunch of functional tests that we need.

tests/Fixtures/project/src/Aspect/DoSomethingAspect.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Go\Tests\TestProject\Aspect;
+
+use Go\Aop\Aspect;
+use Go\Aop\Intercept\MethodInvocation;
+use Go\Lang\Annotation as Pointcut;
```
lisachenko, line 7: 👍 nice import of namespace name

tests/Go/Functional/BaseFunctionalTest.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    protected static $projectDir = __DIR__.'/../../Fixtures/project';
+    protected static $aspectCacheDir = __DIR__.'/../../Fixtures/project/var/cache/aspect';
+    protected static $consolePath = __DIR__.'/../../Fixtures/project/bin/console';
+    protected static $frontControllerPath = __DIR__.'/../../Fixtures/project/web/index.php';
```
lisachenko: I'd prefer to have these assignments aligned... It will be more readable.

Can we also replace these static properties with simple class properties?

tests/Go/Functional/BaseFunctionalTest.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    protected static function warmUp()
+    {
+        return self::exec('cache:warmup:aop');
+    }
+
+    protected static function exec($command, $args = '')
+    {
+        $process = new Process(sprintf('php %s %s %s %s',
```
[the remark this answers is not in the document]
lisachenko: Better to follow PSR here and make it one-liner or extract sprintf part of expression into separate variable.

tests/Go/Functional/BaseFunctionalTest.php
TheCelavi: Not recommended at all. In order to be compatible with static calls, such as `public static function setUpBeforeClass()` they must be static, that is, must be accessible in static context.

Best thing what I can do is to define them as constants, if you prefer that more.

lisachenko:
Oops, merged without squashing changes. Going to clean manually...

lisachenko:
Merged into master as well.

Added `declare(strict_types = 1);` to all new files for master branch.

Issue #340, opened 2017-08-11
[Test] Improve base support for functional testing


TheCelavi:
Class `https://github.com/goaop/framework/blob/2.x/tests/Go/Functional/BaseFunctionalTest.php` should be based class for all functional test, which would improve our testing capability for coverage of various edge cases by simply reproducing other people setup.

In order to have this kind of support, BaseFunctionalTest class and front controller should be modified in order to allow us:

- ~~to define which configuration should be used for AspectKernel for each functional test. That will allow us to test various configuration possibilities, especially useful for those where exception should be thrown and exception message is important.~~ DONE, see [https://github.com/goaop/framework/pull/346](https://github.com/goaop/framework/pull/346)
- ~~to define which aspects should be registered in AspectKernel for each functional test. That will speed up testing process, as well as provide us with some isolation.~~ DONE, see [https://github.com/goaop/framework/pull/346](https://github.com/goaop/framework/pull/346)
- to implement custom `assertWeaved*` methods which would setup base possibility to check weaving of generated, cached classes (of course, without loading, using AST). Partially done
    - ~~`assertClassIsWoven($class)`~~ DONE [https://github.com/goaop/framework/pull/346](https://github.com/goaop/framework/pull/346)
   - ~~`assertClassIsNotWoven($class)`~~ DONE [https://github.com/goaop/framework/pull/346](https://github.com/goaop/framework/pull/346)
   - `assertClassHasStaticMethodJoinPoint($class, $methodName, $joinPoint, $index = null)`
   - `assertClassHasNotStaticMethodJoinPoint($class, $methodName, $joinPoint)`
   - `assertClassHasMethodJoinPoint($class, $methodName, $joinPoint, $index = null)`
   - `assertClassHasNotMethodJoinPoint($class, $methodName, $joinPoint)`

PR #345, opened 2017-08-15
[WIP][Incomplete] Detection of circular reference, part of #340

TheCelavi:
Ok, this is just a proposal of implementing 1. and 2. of issue https://github.com/goaop/framework/issues/340.

Idea is to have a configuration file where you can setup which Kernel implementations should be used and which configuration values should be used.

After that, when you invoke "exec" you can define which configuration should be used. If omitted, `default` will be used.

Configuration key is temporarily set via environment variable in console (GO_AOP_CONFIGURATION) which is read by index.php. When console command is executed, variable is deleted by system, so it is transient. 

@lisachenko What do you think? It is not most elegant solution, but it gives us possibility to choose Kernel (different aspects) and to choose configuration params - so it will be possible to write functional tests where command should fail, or succeed, but with exotic config values.

### DIFF
--- a/src/Exception/CircularWeavingException.php
+++ b/src/Exception/CircularWeavingException.php
@@ -1,0 +1,16 @@
+<?php
+/*
+ * Go! AOP framework
+ *
+ * @copyright Copyright 2014, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Go\Exception;
+
+class CircularWeavingException extends \RuntimeException implements ExceptionInterface
+{
+
+}
--- a/src/Instrument/ClassLoading/CacheWarmer.php
+++ b/src/Instrument/ClassLoading/CacheWarmer.php
@@ -11,6 +11,7 @@
 namespace Go\Instrument\ClassLoading;
 
 use Go\Core\AspectKernel;
+use Go\Exception\CircularWeavingException;
 use Go\Instrument\FileSystem\Enumerator;
 use Go\Instrument\Transformer\FilterInjectorTransformer;
 use Symfony\Component\Console\Output\NullOutput;
@@ -83,6 +84,9 @@ class CacheWarmer
                 );
 
                 $this->output->writeln(sprintf('<fg=green;options=bold>[OK]</>: <comment>%s</comment>', $path));
+            } catch (CircularWeavingException $e) {
+                $this->output->writeln(sprintf('<fg=white;bg=red;options=bold>[ERR]</>: File "%s" is not processed correctly due to detected circular weaving.', $path));
+                throw $e;
             } catch (\Throwable $e) {
                 $displayException($e, $path);
             } catch (\Exception $e) {
--- a/src/Instrument/ClassLoading/SourceTransformingLoader.php
+++ b/src/Instrument/ClassLoading/SourceTransformingLoader.php
@@ -10,6 +10,7 @@
 
 namespace Go\Instrument\ClassLoading;
 
+use Go\Exception\CircularWeavingException;
 use php_user_filter as PhpStreamFilter;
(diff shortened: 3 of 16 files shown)


tests/Go/Functional/BaseFunctionalTest.php
```diff
@@ -22,14 +22,17 @@ protected static function clearCache()
...
+        return self::exec('cache:warmup:aop', '', $configuration);
     }
 
-    protected static function exec($command, $args = '')
+    protected static function exec($command, $args = '', $configuration = null)
     {
-        $commandStatement = sprintf('php %s %s %s %s',
+        $configuration = ($configuration) ? sprintf('GO_AOP_CONFIGURATION=%s ', $configuration) : '';
+
+        $commandStatement = sprintf('%sphp %s %s %s %s',
```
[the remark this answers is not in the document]
TheCelavi: Not applicable.

src/Exception/CircularWeavingException.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @copyright Copyright 2014, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Go\Exception;
+
+class CircularWeavingException extends \RuntimeException implements ExceptionInterface
```
TheCelavi, line 13: I was playing around with https://github.com/goaop/goaop-symfony-bundle/issues/15, because I can easily configure new environment for functional tests, this was easy pick now - I have created new exception class (btw - there are no custom exceptions in library, maybe we should add?)

src/Instrument/ClassLoading/CacheWarmer.php
```diff
@@ -83,6 +84,9 @@ public function warmUp()
                 );
 
                 $this->output->writeln(sprintf('<fg=green;options=bold>[OK]</>: <comment>%s</comment>', $path));
+            } catch (CircularWeavingException $e) {
```
TheCelavi, line 87: Here is the major idea -> when we catch CircularWeavingException -> there is no silencing, we kill the app, there is no continue with execution. 

Display in console, and re-throw exception

src/Instrument/ClassLoading/SourceTransformingLoader.php
```diff
@@ -51,6 +52,13 @@ class SourceTransformingLoader extends PhpStreamFilter
     protected static $filterId;
 
     /**
+     * Transformation log
+     *
+     * @var string
+     */
+    protected static $log = [];
```
TheCelavi, line 59: This is the magic -> we log every transformed class.

src/Instrument/ClassLoading/SourceTransformingLoader.php
```diff
@@ -130,13 +138,21 @@ public static function addTransformer(SourceTransformer $transformer)
...
         $result = null;
+
         foreach (self::$transformers as $transformer) {
             $result = $transformer->transform($metadata);
             if ($result === false) {
                 break;
             }
         }
 
+        if ($result) {
```
TheCelavi, line 149: If file must be transformed, and we already transformed it -> that is because we have circular weaving.

tests/Fixtures/project/src/Application/CircularyWeaved.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Go\Tests\TestProject\Application;
+
+use Go\Tests\TestProject\Annotation as Aop;
+
+class CircularyWeaved
```
TheCelavi, line 7: This is the class that is weaved twice. First is weaved by LoggingAspect, because it is required for CircularWeavingAspect. Then it is matched by CircularWeavingAspect and that is where it fails.

TheCelavi:
Scratch that -> found a better method. Will create new pull request.

PR #347, opened 2017-08-18
[Improvement] Inline definition of advices into classes and beautify formatting

lisachenko:
This PR changes the way of storing advices in proxies, as of now, all advices are stored in the property by default. This should enable opcache optimization for storing this array (it won't be parsed and will be stored in shared memory for classes).

Also, PR beautifies formatting of array and make it clever in proxies, no more ugly `var_export`.


TheCelavi:
Oh this is AWESOME! Now,  it is going to be extremely easy to write planned missing assertions for functional tests (https://github.com/goaop/framework/issues/340). Thanks, you just saved my day :)

lisachenko:
@TheCelavi Wow, surprisingly it helps you! :) 

That case when good change make code better 😄

PR #349, opened 2017-08-18
[WIP] classmap support for composer

lisachenko:
So, it's very dirty-dirty implementation of class map integration with composer. I want to publish this to show implementation and share some thoughts about it.

I'm going to rewrite this one day, but who knows, maybe there is more elegant solution than my :)


lisachenko:
DO NOT MERGE, WIP.

lisachenko:
I'm going to drop this idea. It was good enough to implement though... But now with PHP7.4 and FFI, there is more better way to transform source code directly in the PHP's memory.

PR #351, opened 2017-08-19
[BC BREAK] Simplify interface of IntroductionAdvisor

lisachenko:
This PR introduces small BC break, but no-one should notice that, because it's only implementation detail and not marked with "@api" tag. However code become cleaner and straightforward.

What changed: 
 - Introduction could contain now only one interface/trait per introduction.
 - Method `validateInterfaces()` was removed, because it was unused. This check isn't required at weaving time, we can rely on PHP to check if we implement all methods or should declare class an abstract.
 - Removed `__toString()` from advisor class implementation. Why this method was here?

### DIFF
--- a/src/Aop/Framework/TraitIntroductionInfo.php
+++ b/src/Aop/Framework/TraitIntroductionInfo.php
@@ -19,48 +19,48 @@ class TraitIntroductionInfo implements IntroductionInfo
 {
 
     /**
-     * List of interfaces to introduce
+     * Introduced interface
      *
-     * @var array
+     * @var string
      */
-    private $introducedInterfaces;
+    private $introducedInterface = '';
 
     /**
-     * List of traits to include
+     * Trait to use
      *
-     * @var array
+     * @var string
      */
-    private $introducedTraits;
+    private $introducedTrait = '';
 
     /**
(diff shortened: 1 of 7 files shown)


src/Aop/Framework/TraitIntroductionInfo.php
```diff
@@ -19,48 +19,48 @@ class TraitIntroductionInfo implements IntroductionInfo
...
 
     /**
-     * List of traits to include
+     * Trait to use
      *
-     * @var array
+     * @var string
      */
-    private $introducedTraits;
+    private $introducedTrait = '';
```
TheCelavi, line 33: All seams fine, I just can not figure out why did you set defaults to empty string?

lisachenko: It's preparation for merging into master branch where strict type will be applied to the source code. So, return type hint will be `string` and we should not check if it's null or not.

lisachenko: BTW, you are right, no need to define defaults, because we receive value in the constructor all the time.

lisachenko:
Merged into master as well.

PR #354, opened 2017-08-22
[Feature] Added support for class initializations asserts.

TheCelavi:
Added support for class initializations asserts. Fixed documentation. Minor code improvements.

No BC breaks.


TheCelavi:
Sorry, today is not my day...

lisachenko:
> Sorry, today is not my day...

Hey, mate, everything is ok )  👍  😄

tests/Fixtures/project/src/Aspect/InitializationAspect.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Go\Aop\Aspect;
+use Go\Lang\Annotation as Pointcut;
+
+class InitializationAspect implements Aspect
+{
+    /**
+     * Before class instance initialization.
+     *
+     * @Pointcut\Before("initialization(Go\Tests\TestProject\Application\Main)")
```
lisachenko, line 13: It's fragile joinpoint, maybe we shouldn't use it in framework at all? And just leave it as experimental feature with special configuration?

tests/Fixtures/project/src/Aspect/InitializationAspect.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function beforeInstanceInitialization()
+    {
+        echo 'It invokes before class instance is initialized.';
+    }
+
+    /**
+     * After class initialization.
+     *
+     * @Pointcut\Before("staticinitialization(Go\Tests\TestProject\Application\Main)")
```
lisachenko: Comment name and method name "After" differs from the actual pointcut type, which is "Before"

tests/Go/PhpUnit/ClassMemberNotWovenConstraint.php
```diff
@@ -41,9 +37,13 @@ public function matches($other)
             throw new \InvalidArgumentException(sprintf('Expected instance of "%s", got "%s".', ClassAdvisorIdentifier::class, is_object($other) ? get_class($other) : gettype($other)));
         }
 
-        $wovenAdvisorIdentifiers = $this->getWovenAdvisorIdentifiers($other->getClass());
+        $reflectionClass         = ReflectionClass::create($other->getClass(), $this->configuration);
```
lisachenko: Hmm, what is this? I can't remember `::create()` method in the `ReflectionClass`.

**UPD1** Ah, ok, you have introduced new helper class for that. Maybe rename it to more meaningful name to prevent collision with native `ReflectionClass` and same class from `goaop/parser-reflection`? I would vote for `ProxyClassReflectionHelper` or similar.

tests/Go/PhpUnit/ReflectionClass.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+namespace Go\PhpUnit;
+
+use Go\Instrument\PathResolver;
+use Go\ParserReflection\ReflectionClass as BaseReflectionClass;
+use Go\ParserReflection\ReflectionEngine;
+use Go\ParserReflection\ReflectionFile;
+
+final class ReflectionClass extends BaseReflectionClass
```
lisachenko: Maybe rename it to the `ProxyReflectionClass`? To prevent collision in names with original `ReflectionClass`

tests/Go/Functional/BaseFunctionalTest.php
```diff
@@ -296,4 +296,88 @@ protected function assertPropertyNotWoven($class, $propertyName, $advisorIdentif
...
+    protected function assertClassStaticInitializationWoven($class, $advisorIdentifier = null, $index = null, $message = '')
+    {
+        $identifier = new ClassAdvisorIdentifier(
+            $class,
+            'root',
+            AspectContainer::STATIC_INIT_PREFIX,
+            $advisorIdentifier,
+            $index
+        );
+        $constraint = new ClassMemberWovenConstraint($this->configuration);
```
lisachenko, line 359: 👍  I like how one single constraint class now handles every assertion.

lisachenko (changes requested):

tests/Go/PhpUnit/ClassMemberNotWovenConstraint.php
TheCelavi: Agree, lets make it util class/factory class, that is its purpose.

tests/Fixtures/project/src/Aspect/InitializationAspect.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        echo 'It invokes before class instance is initialized.';
+    }
+
+    /**
+     * Before class initialization.
+     *
+     * @Pointcut\Before("staticinitialization(Go\Tests\TestProject\Application\Main)")
+     */
+    public function afterClassStaticInitialization()
```
lisachenko: What about method name itself and method body? ) It's unmatched with doc block and pointcut

TheCelavi: Something is seriously wrong with my brain today... sorry. I am going to close up for today, no use for me to keep working... :D

lisachenko: Yes, it's always cool to have some time for relax 👍 Don't hurry, this project is maintained for 6 years ) Just keep going slowly

tests/Go/PhpUnit/ProxyClassReflectionHelper.php
```diff
@@ -0,0 +1,44 @@
...
+use Go\ParserReflection\ReflectionFile;
+
+final class ProxyClassReflectionHelper
+{
+    private function __construct()
+    {
+        // noop
+    }
+
+    public static function createReflectionClass($className, array $configuration)
```
lisachenko: Doc block for typehint and IDE?

tests/Go/PhpUnit/ProxyClassReflectionHelper.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+namespace Go\PhpUnit;
+
+use Go\Instrument\PathResolver;
+use Go\ParserReflection\ReflectionClass;
+use Go\ParserReflection\ReflectionEngine;
+use Go\ParserReflection\ReflectionFile;
+
+final class ProxyClassReflectionHelper
```
lisachenko: Missing doc block?

tests/Fixtures/project/src/Aspect/InitializationAspect.php
TheCelavi: I do not know about that, however, I know that if we have it in code, we should have way to test it. Right?

lisachenko (approved):

lisachenko:
Looks good!

lisachenko:
BTW, 
This block:
```php
        if (null === $wovenAdvisorIdentifiers) { // there are no advisor identifiers
            return true;
        }

        if (!isset($wovenAdvisorIdentifiers[$target])) {
            return true;
        }

        if (!isset($wovenAdvisorIdentifiers[$target][$other->getSubject()])) {
            return true;
        }
```

can be replaced with one single check
```php
        if (!isset($wovenAdvisorIdentifiers[$target][$other->getSubject()])) {
            return true;
        }
```

Need to clean this block later in future PR.

PR #356, opened 2017-08-23
Fix incorrect optimization logic for transformers, resolves #355

### LINKED ISSUE #355: [Bug] Object initialization weaving is not working when class in the file hasn't joinpoints
As discussed, if there is no pointcut matching target, transformers are stopped. That leads to the fact that constructor weaving does not occurs in classes where it is being used. 

Source of issue seams to be in: `Go\Instrument\Transformer\CachingTransformer`, https://github.com/goaop/framework/blob/ed7760229190aa5521376ff2dde1d58ede7fbc5a/src/Instrument/Transformer/CachingTransformer.php#L117

I have played around trying to fix the issue: https://github.com/RunOpenCode/framework/blob/353eab93e2312be83aad28b3aeb5d245409b8be7/src/Instrument/Transformer/CachingTransformer.php#L118

I have changed `SourceTransformer`, method `transform()` must return either true or false, void is not an option. I have changed all other transformers in order to follow that signature.

Issue that I have now is that when I run functional test, every class with `new` keyword is weaved and every constructor of every class is being weaved as well. 

So I am missing something big... @lisachenko What I am doing wrong?

lisachenko:
Initially all transformers were included during transformation phase, then I decided to optimize this behaviour for code that should not be woven.

At some point, initialization joinpoint weren't applied to each class even with enabled feature, because `WeavingTransformer` could decide to break transformation if no matching joinpoints are present in the class.

This PR should fix this, but initialization interception still requires transformation of almost every class that uses `new` in the source code.

### DIFF
--- a/src/Instrument/Transformer/SourceTransformer.php
+++ b/src/Instrument/Transformer/SourceTransformer.php
@@ -15,12 +15,26 @@ namespace Go\Instrument\Transformer;
  */
 interface SourceTransformer
 {
+    /**
+     * Transformer decided to stop whole transformation process, all changes should be reverted
+     */
+    const RESULT_ABORTED = 'aborted';
+
+    /**
+     * Transformer voted to abstain transformation, need to process following transformers to get result
+     */
+    const RESULT_ABSTAIN = 'abstain';
+
+    /**
+     * Source code was transformed, can process next transformers if needed
+     */
+    const RESULT_TRANSFORMED = 'transformed';
 
     /**
      * This method may transform the supplied source and return a new replacement for it
      *
      * @param StreamMetaData $metadata Metadata for source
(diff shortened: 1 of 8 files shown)


src/Instrument/Transformer/SourceTransformer.php
```diff
@@ -15,12 +15,26 @@
...
+
+    /**
+     * Transformer voted to abstain transformation, need to process following transformers to get result
+     */
+    const RESULT_ABSTAIN = 'abstain';
+
+    /**
+     * Source code was transformed, can process next transformers if needed
+     */
+    const RESULT_TRANSFORMED = 'transformed';
```
TheCelavi, line 31: 👍  much better solution, much, much better!

PR #357, opened 2017-08-23
Another use of PHP 7 direct invocation

smuggli:


src/Aop/Framework/DynamicClosureMethodInvocation.php
```diff
@@ -38,9 +38,7 @@
     public function proceed()
     {
         if (isset($this->advices[$this->current])) {
-            $currentInterceptor = $this->advices[$this->current++];
-
-            return $currentInterceptor->invoke($this);
+            return ($this->advices[$this->current++])->invoke($this);
```
TheCelavi, line 41: If you ask me, I would prefer more that code has 3 lines, like:

    $this->current++;
    $currentInterceptor = $this->advices[$this->current];		
    return $currentInterceptor->invoke($this);

instead of one liner.

lisachenko (commented):
Hi! Thank you for looking at the code and trying to improve it! 
But it isn't direct closure invocation at all :) , you just call method from an object. And object isn't a `Closure` instance. There are two lines here intentionally to have better debugging.

So, this PR can't be applied.

smuggli:
Wasn't meant to be the same as some of your changes (sorry for the title ... was late) just wanted to lose the ASSIGN ;). For debugging it is definitely better the way it was before.

lisachenko:
@smuggli reason why I changed source code for interceptor classes was that small place is very straightforward and called probably thousands times during one single request and can introduce noticeable delay. That's why for PHP7 I decided to optimize exactly these places to speed up framework performance but leaving code readable. 

Your change is also good (from the point of performance), but debugging and readability are also important.

lisachenko:
@smuggli Don't give up with PR, any help is appreciated! 👍  For example, you could contribute unit tests for primitive pointcuts, for example `AndPointcut`, `OrPointcut`, etc. Or just add more PHP7 typehints in the master branch.

smuggli:
@lisachenko I won't stop because of this ;) Thanks for examples I definitely will give it a shot, but could take some time.

Another thing what do you think about using PHPCsFixer? The `native_function_invocation` rule would give us the opportunity to use the FQN for all native functions. Would improve OPCODES a bit already and maybe more with future PHP releases (like 7.2).

lisachenko:
> I won't stop because of this ;) Thanks for examples I definitely will give it a shot, but could take some time.

👍 Excellent!

> The native_function_invocation rule would give us the opportunity to use the FQN for all native functions. Would improve OPCODES a bit already and maybe more with future PHP releases (like 7.2).

I'd vote against this, even knowing that this can speed up application. This can be applied only to time-critical parts of framework, which affects runtime performance.

Issue #358, opened 2017-08-27
Using annotation always resulting in writing to cache


paventuri:
Given an annotation:
```
/**
 * @Annotation
 * @Target("METHOD")
 */
class MyAnnotation extends Annotation
{
    /**
     * @var string
     */
    public $name = null;

    /**
     * @var string
     */
    public $tag = null;
}
```

Given an aspect:
```
/**
     *
     * @param MethodInvocation $invocation
     *
     * @Around("@execution(MyAnnotation)")
     */
    public function aroundInvocation(MethodInvocation $invocation)
    {

        /** @var MyAnnotation $annotation */
        $annotation = $invocation->getMethod()->getAnnotation(MyAnnotation::class);
        $name = $annotation->name;
    }
```

The problem is that a call to `$invocation->getMethod()->getAnnotation(MyAnnotation::class);` results in a call to `FileCacheReader::getAnnotations`. Which will get the className like so: `$class = $method->getDeclaringClass();`

The declaring class in this case is always `Class__AopProxied` which is not cached. By the way, I'm using `Features::PREBUILT_CACHE` as I have a read-only file systems.

Error:
`The directory "/var/www/html/docroot/bootstrap/cache/aspect/_annotations" is not writable. Both, the webserver and the console user need access. You can manage access rights for multiple users with "chmod +a". If your system does not support this, check out the acl package.`

Is there any workaround for this issue? Thanks

TheCelavi:
As far as I know, pre-built cache means that cache is built first in writable environment and after that deploy in non-writable cache environment (with pre-built cache on).

lisachenko:
Hi! Issue here in `doctrine/annotations` library, which always checks for write access. Upgrading this dependency should solve this issue.

TheCelavi:
@lisachenko Beside that, he said: 

> The declaring class in this case is always Class__AopProxied which is not cached.

Did I understood wrong? When `PREBUILT_CACHE` is on, that means that you have already warmed up cache in writable environment and you are deploying to non-writable environment? So, even with upgrade of `doctrine/annotations`, he should warmup cache prior to deploy and deploy it with cache as well, right?

paventuri:
@TheCelavi - I do warm up the cache before, but `$class = $method->getDeclaringClass();` returns the proxied class and not the original class. The proxied class is not cached as part of the warm up process.

lisachenko:
[...]

Issue #245, opened 2016-02-13
[Insight] Deprecated class usage found - in src/Core/GoAspectContainer.php, line 91


lisachenko:
**[in src/Core/GoAspectContainer.php, line 91](https://github.com/goaop/framework/blob/6c875ffcb4043ccb3f92d384c5e8bc1a21fd5b27/src/Core/GoAspectContainer.php#L91)**

> The `Doctrine\Common\Annotations\FileCacheReader` class has been deprecated in Doctrine Annotations 1.2.3. Use the `Doctrine\Common\Annotations\CachedReader` class instead.

``` php

        $this->share('aspect.annotation.reader', function (Container $container) {
            $options = $container->get('kernel.options');
            $reader  = new AnnotationReader();
            if (!empty($options['cacheDir'])) {
                $reader  = new FileCacheReader(
                    $reader,
                    $options['cacheDir'] . DIRECTORY_SEPARATOR . '_annotations' . DIRECTORY_SEPARATOR,
                    $options['debug']
                );
            }
```

---

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/5998393e-29ea-48f8-8e7e-b19e86daa2db/analyses/72)_

lisachenko:
This issue now results in bug with read-only filesystems and attempts to access annotations, see #358

TheCelavi:
Fixed in https://github.com/goaop/framework/pull/359.

Issue #136, opened 2014-03-17
Interception of internal functions doesn't work if no cacheDir


hax:
Seems a limitation of current implementation.

lisachenko:
You are right, it's impossible to intercept internal functions without an ability to redefine them in a file. There is only one workaround for that is to utilize `php://memory` stream, but I haven't time to check this.

sagikazarmark:
Any update on this?

lisachenko:
Unfortunately, no, because I'm working on [goaop/parser-reflection](https://github.com/goaop/parser-reflection) library now to enable support for PHP5.6 and PHP7 syntax features.

TheCelavi:
I am not 100% sure, but it seams that current implementation can not work without cache directory and weaving classes/function in proxied files.

On top of that, why should we allow engine to work without cache directory at first place? What would be benefits of that?

If there are none, I would modify framework/src/Core/AspectKernel.php, method `normalizeOptions` to insist on cache directory and close this issue.

lisachenko:
Earlier, it was funny to enable AOP without cache dir, it was kind of black magic that was making code working even without cache directory.
Nowadays all parts of framework are optimized for working with cache directory (filemtime checks, opcode cache support, easier debugging of proxies, etc).

We can drop support for empty cache directory, it MUST be present.

@TheCelavi But be aware, that there are read-only filesystems where cache directory is present, but it isn't writable. We should care about this to support modern cloud hostings and read-only file systems.

TheCelavi:
Ok, I will have that in mind when I start on the patch.

TheCelavi:
Cache directory for weaved classes is a "must" for Go! AOP to work. Closing this one on that premise. Changed in https://github.com/goaop/framework/pull/359.

PR #359, opened 2017-08-28
[Fix] Introduce an option to configure cached reader for annotations #136, #358 and #245

TheCelavi:
Fixes for several issues, mostly in regards to caching mechanism. 

Added possibility for users to define custom cache driver for annotations, under option `annotationCache`. Any Doctrine cache can be used.


src/Core/AspectKernel.php
```diff
@@ -199,16 +201,27 @@ protected function getDefaultOptions()
     protected function normalizeOptions(array $options)
     {
         $options = array_replace($this->getDefaultOptions(), $options);
-        if ($options['cacheDir']) {
-            $options['excludePaths'][] = $options['cacheDir'];
+
+        $options['cacheDir'] = PathResolver::realpath($options['cacheDir'], $options['debug']);
```
TheCelavi: In debug mode -> I would check `cacheDir`

src/Core/GoAspectContainer.php
```diff
@@ -91,12 +91,11 @@ public function __construct()
         $this->share('aspect.annotation.reader', function(Container $container) {
             $options = $container->get('kernel.options');
             $reader  = new AnnotationReader();
-            if (!empty($options['cacheDir'])) {
-                $reader = new FileCacheReader(
+            if (!empty($options['annotationCache'])) {
```
TheCelavi: I didn't fixed this one. 

`annotationCache` is not optional, however, current unit tests do not mocks engine configuration, so if I presume that `annotationCache` will always exist here, test fails.

In general - if you want, I can deal with test as well and remove if (!empty(...)) here, however, both code and tests should be revised -> what is required and what is optional...

src/Core/AspectKernel.php
```diff
@@ -216,9 +216,9 @@ protected function normalizeOptions(array $options)
         $options['excludePaths']   = PathResolver::realpath($options['excludePaths']);
 
         if (null === $options['annotationCache']) {
-            $options['annotationCache'] = new PhpFileCache(
+            $options['annotationCache'] = new FilesystemCache(
```
TheCelavi: This is extremely stupid and I don't have any idea why this causing issues...

If `Doctrine\Common\Cache\ PhpFileCache` is used, test fails (and it does not work in general) on PHP 5.x, and 7.0. It works on PHP 7.1 and 7.2.

On the other hand, with same settings, if `Doctrine\Common\Cache\ FilesystemCache` is used, all test passes and it does work.

@lisachenko -> do you have any idea what is happening here? And how to proceed? Should we keep `FilesystemCache` as default implementation, or? Version check and pick accordingly?

BTW I will report issue since it can be easily reproduced...

TheCelavi:
@paventuri In general, it could be implemented by adding some more logic into, per example, caching transformer. However, more logic make us maintain it. Maybe in the future, since, with possibility to choose annotation cache storage this issue may be considered as fixed... I don't know how Alexander feels about it...

README.md
```diff
@@ -196,7 +196,19 @@ Now you are ready to use the power of aspects! Feel free to change anything ever
 
 ### 6. Optional configurations
 
-#### 6.1 Support for weaving Doctrine entities
+#### 6.1 Custom annotation cache
+
+By default, Go! AOP uses `Doctrine\Common\Cache\PhpFileCache` for caching
+annotations. However, should you need to use any other caching engine
```
lisachenko: Is it ok to use "...should you need to use.." here? Maybe replace "should" with "if"?

src/Core/AspectKernel.php
```diff
@@ -199,16 +201,27 @@ protected function getDefaultOptions()
...
-        $options['excludePaths']  = PathResolver::realpath($options['excludePaths']);
+
+        if (!$options['cacheDir']) {
+            throw new \RuntimeException('You need to provide valid cache directory for Go! AOP framework.');
+        }
+
+        $options['excludePaths'][] = $options['cacheDir'];
+        $options['excludePaths'][] = __DIR__ . '/../';
+        $options['appDir']         = PathResolver::realpath($options['appDir']);
+        $options['cacheFileMode']  = (int) $options['cacheFileMode'];
```
lisachenko, line 213: There is a gotcha here, `cacheFileMode` MUST be integer, otherwise strange effects occurs:
```php
echo (int) '0770'; // Outputs 770
echo (int) 0770; // 504, which is correct
```

src/Core/AspectKernel.php
```diff
@@ -199,16 +201,27 @@ protected function getDefaultOptions()
...
+        $options['includePaths']   = PathResolver::realpath($options['includePaths']);
+        $options['excludePaths']   = PathResolver::realpath($options['excludePaths']);
+
+        if (null === $options['annotationCache']) {
+            $options['annotationCache'] = new FilesystemCache(
+                $options['cacheDir'] . DIRECTORY_SEPARATOR . '_annotations' . DIRECTORY_SEPARATOR,
+                '.annotations.cache',
+                0777 & (~$options['cacheFileMode'])
+            );
+        }
```
lisachenko: In my opinion this should be done in `GoAopContainer` class, kernel should not work with such classes directly.

src/Core/GoAspectContainer.php
lisachenko: It's ok to check with `!empty($options['annotationCache'])` here.

src/Core/AspectKernel.php
lisachenko: I can propose several solutions: 
1) To extend `GoAopContainer` class and to redefine definition of annotation.reader class.
2) Make annotation cache an option that accepts a closure that should return instance of annotation reader and then just store it in the container if declared. By default, `GoAopContainer` will define a closure that returns simple `FilesystemCache` cached reader.

lisachenko (changes requested):

README.md
TheCelavi: Mission impossible? "Your mission is, should you choose to accept it, ..." ? :D

lisachenko: For my English knowledge it is strange phrase with "should" usage )) But we can ignore my comment. 

My version is: "However if you need to use any other caching engine...". More straightforward ) But I can not be sure if your version isn't correct )

src/Core/GoAspectContainer.php
```diff
@@ -88,19 +89,35 @@ public function __construct()
...
+            if (!empty($options['cacheDir'])) {
+                return new DoctrineCache\FilesystemCache(
+                    $options['cacheDir'] . DIRECTORY_SEPARATOR . '_annotations' . DIRECTORY_SEPARATOR,
+                    '.annotations.cache',
+                    0777 & (~$options['cacheFileMode'])
                 );
             }
 
-            return $reader;
+            return new DoctrineCache\ArrayCache();
```
TheCelavi, line 107: This can occur when running unit tests, so I fallback to ArrayCache in order for tests to pass.

src/Core/GoAspectContainer.php
```diff
@@ -88,19 +89,35 @@ public function __construct()
             );
         });
 
-        $this->share('aspect.annotation.reader', function(Container $container) {
+        $this->share('aspect.annotation.cache', function(Container $container) {
```
TheCelavi, line 92: Ok, so moving to container - agree, much better. Defined new service, `aspect.annotation.cache`, so reader can get that service for cache engine for annotations. 

Much better, thanks.

lisachenko (approved):
Looks good now

paventuri:
Is this going to be merged into goaop:1.X too?

lisachenko:
@paventuri unfortunately, no, versions 1.x are no maintained anymore. Consider to upgrade to 2.x

PR #361, opened 2017-09-02
Added missing command to console commands, added missing dependency. 

TheCelavi:


TheCelavi:
Sorry about the mess.

Just figure out that we had command missing in console, while new cache implementation for annotation requires from us to define additional dependency.

lisachenko (approved):

lisachenko:
Merged into 2.x and master

PR #362, opened 2017-09-03
[Feature] Implementation of AST+token modification instead of direct source transforming

lisachenko:
This PR changes the logic of transformers, they can use AST now and operate directly on token stream using token positions from AST to filter interesting places.

### DIFF
--- a/src/Instrument/Transformer/FilterInjectorTransformer.php
+++ b/src/Instrument/Transformer/FilterInjectorTransformer.php
@@ -13,6 +13,8 @@ namespace Go\Instrument\Transformer;
 use Go\Core\AspectKernel;
 use Go\Instrument\PathResolver;
 use Go\Instrument\ClassLoading\CachePathManager;
+use PhpParser\Node\Expr\Include_;
+use PhpParser\NodeTraverser;
 
 /**
  * Transformer that injects source filter for "require" and "include" operations
@@ -97,7 +99,7 @@ class FilterInjectorTransformer implements SourceTransformer
         }
 
         $resource = (string) $originalResource;
-        if ($resource['0'] !== '/') {
+        if ($resource[0] !== '/') {
             $shouldCheckExistence = true;
             $resource
                 =  PathResolver::realpath($resource, $shouldCheckExistence)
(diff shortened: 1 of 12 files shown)


lisachenko:
@TheCelavi it's working solution for AST-based transformers, please look at it if you have time ) Your review is appreciated

src/Instrument/Transformer/FilterInjectorTransformer.php
```diff
@@ -122,66 +124,32 @@ public static function rewrite($originalResource, $originalDir = '')
      */
     public function transform(StreamMetaData $metadata)
     {
-        if ((strpos($metadata->source, 'include') === false) && (strpos($metadata->source, 'require') === false)) {
+        $includeExpressionFinder = new NodeFinderVisitor([Include_::class]);
+
+        // TODO: move this logic into walkSyntaxTree(Visitor $nodeVistor) method
+        $traverser = new NodeTraverser();
```
TheCelavi, line 130: Have you think about to maybe have:

    $nodes = NodeFinderVisitor::getInstance()->findNodes([]);

? `NodeFinderVisitor` could also be a `final` class and annotated as internal...

While we are at final classes - you should start to think about making majority of classes final and annotated as internal - it will be much easier to introduce changes without BCs.

TheCelavi:
I think that there is a hidden problem here coming from the fact that token stream gets changed, but AST does not reflects those changes. 

That means that in chain of transformers, first transformer could change code, however, second one will introspect AST of non-transformed code and modify already modified token stream. AST would, in that case, contain `startTokenPos` and `endTokenPos` with wrong indexes, right?

I am wrong?

src/Instrument/Transformer/FilterInjectorTransformer.php
lisachenko: I like the idea of making classes final and providing API for public part. Likely, AOP is low coupled and provided only small API via `Joinpoint` object instances, all this classes can already tagged via `@api` hash tag and part of them are final.

What about annotating methods as internal: in IDE all methods will be displayed striked like this ~~->getFoundNodes();~~ and this will add false sense that methods are deprecated.

I have a willing to rewrite transformers for 3.0 and break BC for that. What I want is to use `goaop/ast-transformer` library that will contain simple node visitors (`PhpParser\NodeVisitor` instances) and instead of scanning nodes in each transformer, just do this in one cycle. `goaop/ast-transformer` should also provide API for modification of original source code, hiding all details about token positions, updating of AST during transformation, etc...

lisachenko:
@TheCelavi regarding your last question: you are right. At that moment implementation is straightforward and fragile, because changes should be performed on token stream.

To keep positions of tokens in AST there are some rules now:
- it's safe to add some content into token value (e.g. $token[1] .= 'Some content')
- it's safe to unset one or more tokens by their offset from AST
- do not add new tokens or adjust their offsets, information in AST won't be updated and this will lead to wrong result.

I don't like this solution fully, but it's small step towards to node visitors/transformers that should be implemented in `goaop/ast-transformer` library. However code of transformers are more readable and reliable with AST right now.

src/Instrument/Transformer/FilterInjectorTransformer.php
TheCelavi: 👍

TheCelavi:
It is not possible re-create AST from modified token stream or you believe that it is too expensive operation? 

BTW - you have small merge conflict which should be resolved in order to merge this.

lisachenko:
> It is not possible re-create AST from modified token stream or you believe that it is too expensive operation?

It's possible, but I don't think that it will be nice idea, we should work on original file content. Otherwise, line numbers won't be same and debugger will show incorrect places in IDE.

TheCelavi:
> Otherwise, line numbers won't be same and debugger will show incorrect places in IDE.

How is that possible? Let say that we have code X, and there are transformers A, B, C.

Each transformer transforms code X in such way that code formatting is preserved (line numbers are the same).

X => transformed by A => X' , formatting preserved
X => transformed by B => X' , formatting preserved
X => transformed by C => X' , formatting preserved

X => transformed by A => transformed by B => transformed by C => X''' => formatting should be preserved as well, right?

So, if you do any transformation of token streams, and you regenerate AST from that token stream, I do not see how it is possible to mess up code formatting...?

lisachenko:
> So, if you do any transformation of token streams, and you regenerate AST from that token stream, I do not see how it is possible to mess up code formatting...?

What if some transformer unset token for line? Then formatting could be broken, even PHP syntax can be invalid (just imagine that you remove `T_CLASS` token). So it's very dangerous to work directly on tokens, because it can result in errors.

So `tokens => AST => tokens => AST => tokens => source code` is bad for goaop framework itself. What I want is:
```
    AST     =>     AST
   /   \          /   \
Tokens Tokens  Tokens Tokens
```

We should work with AST only and it transforms underlying token stream according to required changes. And API should be readable like `makeFinal()` or `removeModifiers([T_PRIVATE, T_PUBLIC])`, etc..

TheCelavi:
> just imagine that you remove T_CLASS 

Yeah, but we are not going to allow that kind of transformer in  codebase, right? :)

What I am implying here is that you are traverse AST, and based on info from AST you are modifying token stream. However, updated token stream will not reflect AST after transformation.

If we got information that there are transformers A and B, where A will transform code X to X' based on AST, and B would transform code X' to X'', but not X (B depends on transformation of A), not having AST that reflects transformed code will in that case cause an issue.

I am not sure if now we can get into situation? Maybe with initialization interception and `__FILE__` interception in same class? Example:

```
class Example 
{
    public function dummyMethod()
    {
         require_once __FILE__.'/Intercepted.php';
         new Intercepted();  // <-- this initialization will be intercepted
    }
}
```

Is this going to be weaved properly ?

TheCelavi:
What I was thinking, in the future is, per example, class replacement transformer (which we do not have, but it is very, very possible to implement):

-> Find all `\ReflectionClass` and replace with `\My\ReflectionClass` in source code

and then other transformer can kick in:

-> intercept initialization of all `\My\ReflectionClass` in source code

if AST is not updated to modified token stream, successive weaving is not possible.

We do not have that now, of course, so this can be merged.

lisachenko:
Generally, this implementation is working, but it's not ideal. We can investigate possible ways how to modify this later to make working complex transformations like in your previous example.

Anyway, it will be AST-based approach ) So, if you don't have serious objections against this, we can merge this PR. Hopefully, I didn't break anything for complex cases...

TheCelavi:
No, I don't have. In general, if you do not have time, I will create functional test for complex, multi-transformer weaving. Just to check and to confirm that what is supported now, it will work, like previous example:

```
class Example 
{
    public function dummyMethod()
    {
         require_once __FILE__.'/Intercepted.php'; // <- this statement will be intercepted
         new Intercepted();  // <-- this initialization will be intercepted
    }
}
```

If I find an issue, I will notify you, of course.

lisachenko:
Ok, thank you! Functional tests would be cool 👍

lisachenko:
Merged into master as well.

Issue #368, opened 2017-11-08
Typehints in global namespace


daniel-sc:
It seems there is a bug with typehints in global namespace.
I documented this with the following commit https://github.com/daniel-sc/framework/commit/4439c4d445ad151560a0fc5dfa3ede0629c7edc5 on the branch https://github.com/daniel-sc/framework/tree/bug-typehint

The test fails, because the woven class contains the following problematic typehint:
```php
public function publicMethodFixedArguments(\\Exception $a, $b, $c = NULL)
```

This leads to errors in reality like:
```
ParseError: syntax error, unexpected '\' (T_NS_SEPARATOR), expecting identifier (T_STRING)
```

lisachenko:
Hi, @daniel-sc!

Thank you for reporting this, looks like there should be ltrim here for normalizing global namespace.

daniel-sc:
Hi @lisachenko 

I updated the PR accordingly.

lisachenko:
Closed in #369

PR #369, opened 2017-11-08
Fix for typehint bug

daniel-sc:
Test and fix for #368


src/Proxy/AbstractProxy.php
```diff
@@ -109,6 +109,10 @@ protected function getParameterCode(ReflectionParameter $parameter): string
             $nullablePrefix = (PHP_VERSION_ID >= 70100 && $reflectionType->allowsNull()) ? '?' : '';
             $nsPrefix       = $reflectionType->isBuiltin() ? '' : '\\';
             $type           = $nullablePrefix . $nsPrefix . (string) $reflectionType;
+            // prevent double "\\" prefix for global namespace type hints:
```
TheCelavi: @lisachenko Other option would be to check for http://php.net/manual/en/reflectionclass.isinternal.php on line 110 as well. Kinda don't like string manipulation here...

tests/Go/Instrument/Transformer/_files/class-typehint.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+class TestClassTypehint {
+
+    public function publicMethodFixedArguments(Exception $a, $b, $c = null) {}
```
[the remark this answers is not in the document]
TheCelavi: @daniel-sc What happens when Exception is NS prefixed? Like this:

     public function publicMethodFixedArguments(\Exception $a, $b, $c = null) {}

Same issue occurs?

src/Proxy/AbstractProxy.php
daniel-sc: I don't like the dump string manipulation either, but am not really understanding in which cases we need to add the `\` ..
Could you show how this could be done?

lisachenko: Maybe just add `ltrim()` code here to prevent duplication?

tests/Go/Instrument/Transformer/_files/class-typehint-woven.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+class TestClassTypehint__AopProxied {
```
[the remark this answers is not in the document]
lisachenko: Need to shut up these notices from Nitpick for test directories...

src/Proxy/AbstractProxy.php
TheCelavi: Ltrim FTW !

daniel-sc:
@lisachenko @TheCelavi Are there any other things you'd need to get updated to merge this PR?

TheCelavi (approved):

TheCelavi:
Sorry, I guess we are both swamped with our own work... end of the year, you know....

Either way, thank you much for report and bugfix!

lisachenko:
My apologize to @daniel-sc! Thank you for contribution! And one more thank to @TheCelavi for merging this PR!

Nice to have additional hands on project )

daniel-sc:
No problem - thanks for merging!

PR #373, opened 2017-11-26
[Feature] interception of final methods, resolves #372

### LINKED ISSUE #372: [Feature] Interception of final methods
Currently, framework doesn't support interception of final methods, but after switching to the AST-base transformers, it can be easily solved by asking offsets for the methods.

Idea of intercepting final methods is following.
We take original class `Foo` and make it's intercepted final method `bar` non-final. Then create child with this method and make it final to preserve final logic.

```php
final class Foo{
    final public function bar() {}
}
```
Will be converted to the:
```php
class Foo__AopProxied{
    public function bar() {}
}

final class Foo extends Foo__AopProxied {
    final public function bar() {}
}
```

lisachenko:
This PR enables interception of final methods in the classes, see #372 for additional details about implementation.


TheCelavi (approved):
👍

Issue #348, opened 2017-08-18
[Support] AOP is not applied to the static method called with self:: keyword


dulio:
main code:
```
class TestService {
    public static function testOne($a, $b) {
        return self::testTwo($a , $b);
    }

    /**
     * @Cacheable(ttl=60)
     */
    public static function testTwo($a, $b) {
        return $a + $b;
    }
}
```

call code:
`$res = TestService::testOne(11, 4);`

I found that: the second called static function not work with AOP
but the first static function works well.

go aop version: 2.1.2

TheCelavi:
We can not see the issue here, you have to provide us with full example, how did you imported Cacheable annotation and how Aspect looks like - there are too few details here to even grasp what is wrong. Beside that, what is the expected result and how did you came to conclusion that function is not woven?

lisachenko:
I know, what the root cause here.

In order to use AOP with static methods, you should use `static` keyword in methods body to enable Late Static Binding (aka LSB)

So, change 
```php
return self::testTwo($a , $b);
```
to the
```php
return static::testTwo($a , $b);
```

TheCelavi:
@lisachenko Wow, that is quite important info! Will add it do documentation as well.

dulio:
I got it, thx!

lisachenko:
@dulio You are welcome! Thank you for using the framework!

TheCelavi:
After some thinking -> this is a bug, not framework limitation. We can fix this.  Consider this:

    class A {
        static function x() {    
        }

        static function y() { 
            self::x();
        }
    }

if we weave class A on function x, we will get:

    class A_AopProxied {
        static function x() {    
        }

        static function y() { 
            self::x();
        }
    }

    class A {
        private static $__joinPoints = []

        static function x() {    
            return self::$__joinPoints['static:x']->__invoke($this);
        }
    }

and that is why `static` has to be used - **unless!!!** when we generate `A_AopProxied` do simple string replacement where `self::` is replaced with `A::` then we will get:

    class A_AopProxied {
        static function x() {    
        }

        static function y() { 
            A::x();
        }
    }

    class A {
        private static $__joinPoints = []

        static function x() {    
            return self::$__joinPoints['static:x']->__invoke($this);
        }
    }

and there will be no need for late static binding. Right?

TheCelavi:
Two reason why I propose this:
[...]

Issue #139, opened 2014-03-28
 Catchable fatal error


mustela:
I coded an aspect to work with this class (well I just copied it from your examples :) )

```
namespace Maven\Front\Actions;

class AddToCartAction extends Action {

    public function __construct ( \Maven\Front\Step $step, \Maven\Core\Cart $cart, $data ) {
        parent::__construct( $step, $cart, $data );
    }

    /**
     * 
     * @return \Maven\Core\Message
     * @throws \Maven\Exceptions\InvalidObjectTypeException
     */
    public function execute () {
         ........
    }
}

```

But when it runs I get 

Catchable fatal error: Argument 2 passed to Maven\Front\Actions\AddToCartAction::__construct() must be an instance of Maven\Core\Cart, instance of Maven\Core\Cart__AopProxied given, called in ....

This is the aspect

```
/**
 * Monitor aspect
 */
class MonitorAspect implements Aspect
{

    /**
     * @param MethodInvocation $invocation Invocation
     * @Before("execution(public Maven\Core\Cart->*(*))")
     */
    public function beforeMethodExecution(MethodInvocation $invocation)
    {

        $obj = $invocation->getThis();
        echo 'Calling Before Interceptor for method: ',
             is_object($obj) ? get_class($obj) : $obj,
             $invocation->getMethod()->isStatic() ? '::' : '->',
             $invocation->getMethod()->getName(),
             '()',
             ' with arguments: ',
             json_encode($invocation->getArguments()),
             "<br>\n";

    }

}
```

Any clue?

lisachenko:
This issue is related to https://github.com/lisachenko/go-aop-php/issues/94. Please, show me the code, where instance of `Maven\Core\Cart` is created. Due to the nature of AOP implementation it requires Late Static Binding enabled for class creation. Probably, you call `new self()` inside `Maven\Core\Cart` or some static method with `self::methodName()`?. To work with AOP, you should change this to `static` keyword: `new static()`, `static::methodName()`, etc... Alternative way: exclude factory methods from the pointcut.

mustela:
Yes, you were right. I just changed the new self() for new static(), and it's working now. 

Thanks again!

lisachenko:
@mustela Welcome!

Issue #130, opened 2014-02-12
Issue loading class by static function using "self"


eriknyk:
Hi, 
I found a issue doing the following:
I create a class:  Example.php

```
<?php
class Example
{
    public function hello($value='')
    {
        echo "hello $value" . PHP_EOL;
    }
    public static function load()
    {
        $me = new self();
        return $me;
    }
}
```

and I'm setting a LoggerAspect using:

```
@Before("execution(public Example->*(*))")
```

after setting my: Kernel\ApplicationAspectKernel::getInstance(),.. etc.
and execute 

```
$e = Example::load();
$e->hello("ERIK");
```

It doesn't works!!! :(

But if I change my static funcion load(), so:

```
public static function load()
{
    $me = new Example(); // <-- Changing from: new self() to: new Example()
    return $me;
}
```

It works :), But I'm thinking that it is a unexpected behavior,.. can you tell me, it is a bug, or a limitation please,..

Best Regards.

lisachenko:
This issue is also limitation, not a bug, because AOP creates a reverse-proxy class by renaming original class and extending it, so if you want to receive normal instance, you should use Late Static Bindind, just replace `new self()` with `new static()`.

Duplicates https://github.com/lisachenko/go-aop-php/issues/94

lisachenko:
It will be nice to add this issue to the documentation, so everyone will be able to notice it.

Issue #94, opened 2013-08-29
When proxy classes trick everyone


DavertMik:
I have a code:

``` php
<?php
UserService::activateUser(User $user) { }
?>
```

and 

``` php
<?php
class User {
    function register() {
        UserService::activateUser($this);
     }

}
?>
```

`User::activateUser` receives `User_AopProxied` instead of `User`, thus code fails.

Do you have any ideas or workarounds for this?

lisachenko:
If you intercept methods in the  `User` class then `$this` should point to the proxy class with `User` name and typehint should work correctly. But this issue is typically happen with expressions such as `new self()` when a source code doesn't aware of possible children classes.

lisachenko:
This issue is related to https://github.com/symfony/symfony/pull/8367

PR #381, opened 2017-12-25
Fix handling of special type hints, resolves #376

### LINKED ISSUE #376: [Bug] Incorrect handling of special type hints
Generated proxies don't support special typehints like `self`, `parent` because they are not builtin and are prefixed with leading slash like this: `\self`

Need to put them into exclusion list.

lisachenko:
This PR should fix handling of `self` and `parent` as type hints.


lisachenko:
Implementation is wrong... My fault, need to rework this PR.

lisachenko:
Hmm, this issue is very tricky. I want definitely to drop support for `parent` typehint, because calculation of parent name with proxy generation looks weird. And actually no-one uses `parent` as typehint, so it would be natural choice to drop this stuff.

I also have troubles with handling `self` in PHP5.6, because there is no way to query such information via reflection API. So I have big wish not to support this typehint for PHP5.6 and fix this issue only for master for PHP>=7.0. 

Even with PHP>=7.0 handling of `self` keyword requires knowledge of context, for example when grandparent has method with `self` as parameter typehint and we intercepting this method in child, then we should resolve what value should we use: should it be grandparent class or grandparent proxy (if method was intercepted in grandparent too).

TheCelavi:
```php
class A_AopProxied 
{
    const X = 'value'

    public function foo($var = self::X) {}
    public function bar($var = self::class) {}
}

class A extends A_AopProxied
{
    public function foo($var = self::X) {  /* weaved */ }
    public function bar($var = self::class) { /* weaved */ }
}
```

I am trying to understand your dilema, is this the example that you are referring?

When using constant (e.g. `self::X`) seams that nothing should be resolved, it will work either way.

When using class name, well, proxied class should resolve to child class name, that is, `self::class` should resolve in, in our example, `A::class`.

If you are referring to this kind of type-hinting:

```php
class A
{
    public function foo(self $var) {}
}
``` 
My question is - does anyone ever write code like that? That is like really, really edge case (BS if you ask me). Either way, it seams like it should resolve to, as in our example, to A::class.

Or am I getting something wrong here?

lisachenko:
@TheCelavi `self` typehint is typically used for return values to have fluent interface contract.  But dilemma with `self` is just combination of inheritance and interception rules.

So, let's start with following hierarchy:
```php
class A {
    public function setBar(): self {}
}

class B extends A {
    public function setBaz(): self {}
}
```

Then we want to intercept this method `setBaz()` from the class `B`. During proxy generation I will have full static reflection about class `B`, but I also need to rename this class to `B__AopProxied` and then I need to create proxy, keeping the signature for original method. This requires replacing `self` typehint with `B__AopProxied` FQN.

Hierarchy will be as following:
```php
class A {
    public function setBar(): self {}
}

class B__AopProxied extends A {
    public function setBaz(): self {}
}

class B extends B__AopProxied {
    public function setBaz(): B__AopProxied {}
}
```

But if I want to intercept method `setBar()` from the parent class `A` then generated code should refer to the class `A`, not to the proxied class:
```php
class A {
    public function setBar(): self {}
}

class B__AopProxied extends A {
    public function setBaz(): self {}
}

class B extends B__AopProxied {
    public function setBar(): A {}
    public function setBaz(): B__AopProxied {}
}
```

And finally, if this method `setBar()` will be intercepted in the both classes, then code should be generated like this.
```php
class A__AopProxied {
    public function setBar(): self {}
}

class A extends A__AopProxied {
    public function setBar(): A__AopProxied {}
}

class B__AopProxied extends A {
    public function setBaz(): self {}
}

class B extends B__AopProxied {
    public function setBar(): A__AopProxied {}
    public function setBaz(): B__AopProxied {}
}
```
All these rules for handling `self` keyword produce a lot of pain :) Summarize, to resolve `self` typehint I should check where it was declared, was it intercepted in parent or in child class and replace appropriately in generated proxy

TheCelavi:
Thanks, I totally forgot about return types and type-hinting of those...

lisachenko:
Same applied to parameter typehint:

```php
class A {
    public static function adjustInstance(self $instance) {}
}
```

lisachenko:
What if replace `self`keyword with resolved value in an original class in advance?
```php
class A {
    public function setBar(): self {}
}

class B__AopProxied extends A {
    public function setBaz(): B {} // replaced `self` with FQN
}

class B extends B__AopProxied {
    public function setBaz(): B {} // just FQN from parent class, should work
}
```

TheCelavi:
https://en.wikipedia.org/wiki/Occam%27s_razor

lisachenko:
Ok, let's try to do this from another side, hope this won't be very complex :)

lisachenko:
Just want to say: "Wow, PHP...", there are huge amount of places where `self` can appear.

1. Parameter type for method e.g. `public function foo(self $instance)`
2. Parameter type for closure, defined in the class: `$func = function (self $instance) {}`
3. Return type for method or closure in the class: `public function foo(): self {}`
4. Static method call: `self::someMethod()`
5. Class constant fetch: `$class = self::class; $const = self::SOME_CONSTANT;`
6. New instance creation: `new self();`
7. Instance of check: `$instance instanceof self`
8. Catch section (if `self` refers to instance of Exception): `catch (self $exception)`
9. Static property fetch: `$value = self::$someStaticVariable;`

I have implemented visitor that replaces `self` with FQN, but then I've discovered that 4) and 9) can fail for private properties and methods, because child class scope won't contain private properties and methods.

Interesting fact, that if you have any private static method, then you could always call it from the child class scope:

```php
class A {
    public function test() {
        echo B::bar();
    }

    private static function bar() {
        echo 'Yes';
    }
}

class B extends A {
}

$a = new B;
$a->test();
```
So for "method call" replacing `self` with child class name will work until child class try to declare the same private method, then everything will be broken.

For static property fetching results are more predictable: it's impossible to replace `self` with child class, because if property was declared as private then fatal will be raised:

```php
class A {
    private static $foo = 42;
    public function test() {
        echo B::$foo;
    }
}

class B extends A {
}

$a = new B;
$a->test();
// Fatal error: Uncaught Error: Cannot access property B::$foo
```
For this case it will be easier to leave `self::$someProperty` in code as is and do not transform such nodes. Summarize everything 1)-8) are implemented and 9) should not be transformed.

PR will be updated soon...

lisachenko:
Implemented via AST-transformation of original class

This also resolves old issues #348, #139,  #130, #94, because `self` is replaced with resolved class name now.

PR #388, opened 2018-01-22
[Feature] Implementation of annotation getters for the ClassField joinpoint (#387)

### LINKED ISSUE #387: [Support] Accessing annotations of a property for the ClassField
Hello !

I want to annotate some properties to set a defaut value. I made an advice which work well to catch access to annotated properties but I can't access to these annotations once I'm there.

My annotation :
```php
namespace AZE\aop\annotation;

use Doctrine\Common\Annotations\Annotation;

/**
 * @Annotation
 *
 * @Target("ALL")
 */
class Config extends Annotation
{
    public $instance;
}
```

My annotated class :
```php
namespace data\config;

use AZE\aop\annotation\Config;

class Configuration
{
    /**
     * @Config("application")
     */
    public $application = "application_set";

    /**
     * @Config("my.custom.value")
     */
    public $value = "value_set";

    public $model = "aze";
}
```

```php
namespace AZE\aop\aspect;

use Go\Aop\Aspect;
use Go\Aop\Framework\ClassFieldAccess;
use Go\Lang\Annotation\Around;

class ConfigAspect implements Aspect
{
    /**
     * @Around("@access(AZE\aop\annotation\Config)")
     */
    public function getConfigValueAround(ClassFieldAccess $field)
    {
        var_dump($this);
        var_dump($field);
        return "around";
    }
}
```

I don't see any possibility to access current annotations of the property once I'm in my advice. Is it possible ?

Thanks :)

TheCelavi:
- Allows direct access to class property annotations from aspect
- Added tests for Go\Aop\Framework\ClassField 
- Added test that verifies that access to annotations exists. 

NOTE: Due to singleton implementation of Kernel and access to Reader is done via service locator pattern and static access, it is hard to do a proper testing via injecting mock.

But, at least these tests would detect missing presence of methods which will warn us that we are introducing BC.

### DIFF
--- a/src/Aop/Support/AnnotatedReflectionProperty.php
+++ b/src/Aop/Support/AnnotatedReflectionProperty.php
@@ -1,0 +1,60 @@
+<?php
+declare(strict_types = 1);
+/*
+ * Go! AOP framework
+ *
+ * @copyright Copyright 2014, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Go\Aop\Support;
+
+use Doctrine\Common\Annotations\Reader;
+use Go\Core\AspectKernel;
+use ReflectionProperty;
+
+/**
+ * Extended version of ReflectionProperty with annotation support
+ */
+class AnnotatedReflectionProperty extends ReflectionProperty
+{
+    /**
+     * Annotation reader
+     *
+     * @var Reader
+     */
+    private static $annotationReader;
+
+    /**
+     * Gets a property annotation.
+     *
+     * @param string $annotationName The name of the annotation.
+     * @return mixed The Annotation or NULL, if the requested annotation does not exist.
+     */
+    public function getAnnotation(string $annotationName)
+    {
+        return self::getReader()->getPropertyAnnotation($this, $annotationName);
+    }
+
+    /**
+     * Gets the annotations applied to a property.
+     */
+    public function getAnnotations(): array
+    {
+        return self::getReader()->getPropertyAnnotations($this);
+    }
+
+    /**
+     * Returns an annotation reader
+     */
+    private static function getReader(): Reader
+    {
+        if (!self::$annotationReader) {
+            self::$annotationReader = AspectKernel::getInstance()->getContainer()->get('aspect.annotation.reader');
+        }
+
+        return self::$annotationReader;
+    }
+}
--- a/src/Aop/Support/AnnotatedReflectionProperty.php
+++ b/src/Aop/Support/AnnotatedReflectionProperty.php
@@ -18,7 +18,7 @@ use ReflectionProperty;
 /**
  * Extended version of ReflectionProperty with annotation support
  */
-class AnnotatedReflectionProperty extends ReflectionProperty
+class AnnotatedReflectionProperty extends ReflectionProperty implements AnnotationAccess
 {
     /**
      * Annotation reader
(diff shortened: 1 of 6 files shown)


TheCelavi:
Instead of those nasty tests, this is much better, we can introduce interface that should provide access to annotations. Much cleaner.

lisachenko:
Thanks for PR )

Should we drop support for 2.x? And concentrate support only on new 3.x (aka master)?

src/Aop/Support/AnnotatedReflectionProperty.php
```diff
@@ -0,0 +1,60 @@
+<?php
+declare(strict_types = 1);
+/*
+ * Go! AOP framework
+ *
+ * @copyright Copyright 2014, Lisachenko Alexander <lisachenko.it@gmail.com>
```
lisachenko, line 6: Year isn't correct :)

src/Aop/Support/AnnotationAccess.php
```diff
@@ -0,0 +1,31 @@
...
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Go\Aop\Support;
+
+/**
+ * Provides access to annotations from reflection class/property/method.
+ */
+interface AnnotationAccess
```
lisachenko, line 17: 👍  for interface extraction refactoring

src/Aop/Support/AnnotationAccess.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Gets a annotation.
```
lisachenko: Typo: `aN annotation`, also in the child class

tests/Go/Aop/Framework/ClassFieldAccessTest.php
```diff
@@ -0,0 +1,35 @@
+<?php
+declare(strict_types = 1);
+
+namespace Go\Aop\Framework;
+
+use Go\Aop\Support\AnnotationAccess;
+
+class ClassFieldAccessTest extends \PHPUnit_Framework_TestCase
```
lisachenko, line 8: 👍  for the test case

lisachenko (changes requested):
Need to fix years in file docBlocks, all other stuff is correct

lisachenko:
AppVeyor is not working, crappy Windows shit... Couldn't install PHP7.2.1 from Chokolatey recipe due to unknown bug with KB packet.

src/Aop/Support/AnnotatedReflectionProperty.php
TheCelavi: I did copy/paste of docBlock.

Different classes have different years, I checked.

To be honest - I am ok for having author and/or copyright annotation, but year in docblock does not have any particular value, we could just remove it - it poses only unnecessary maintenance.

TheCelavi:
> Should we drop support for 2.x? And concentrate support only on new 3.x (aka master)?

PHP 5.x seams dead/dying, and it is easier for us to maintain only one version.

src/Aop/Support/AnnotatedReflectionProperty.php
lisachenko: > To be honest - I am ok for having author and/or copyright annotation, but year in docblock does not have any particular value, we could just remove it - it poses only unnecessary maintenance.

Looks reasonable. Accepted, will do this for the master branch.

lisachenko (approved):

lisachenko:
Merged into master, thanks!

PR #389, opened 2018-01-24
Correction on demo to show source code of class

Airmanbzh:
Correction on demo to show the content of the tested class

Before : 
goaop_error

After :
goaop_no-error

PS : it's my first (little) contribution to goaop so I may have miss some point to test or to verify

### DIFF
--- a/demos/index.php
+++ b/demos/index.php
@@ -260,7 +260,16 @@ if ($example):
       <div class="panel-body well panel-collapse collapse out" id="collapseTwo">
 <?php
 $refObject = new ReflectionObject($example);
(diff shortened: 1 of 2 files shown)


lisachenko:
Ah, it's small forgotten code :) Nice catch!

But it shouldn't be fixed in the `MagicConstantTransformer` class.

It should be fixed directly in the demo code, because it shouldn't be used for generated proxies, it's just for the demo.

Could you please, update only demo code?

Airmanbzh:
I updated the code to do this action in the "index.php" file

lisachenko (approved):

lisachenko:
Thanks!

PR #393, opened 2018-04-07
[Improvement] Decompose proxy generator

lisachenko:
This PR adds one more dependency, but makes core more readable and maintainable. Hope, that `zendframework/zend-code` is good enough to be used as code generator.


TheCelavi (commented):
Didn't went into every detail, but seams legit. I like zend-code, why not?

lisachenko:
@TheCelavi Thanks for reviewing! 
Only the disadvantage is that `zendframework/zend-code` has strange dependency on `zendframework/zend-eventmanager` (see https://github.com/zendframework/zend-code/issues/32#issuecomment-357039345) and `zendframework/zend-code` is not maintained last years (last commit was on `20 Oct 2017`). 

But all @Ocramius's tools depend on `zendframework/zend-code`, so this should be stable enough.

TheCelavi:
> But all @Ocramius's tools depend on zendframework/zend-code, so this should be stable enough.

That means that you are betting on Ocramius's good health and well being? He is old, and I heard that he drinks a lot...  :D :D :D

Ocramius:
Don't forget that I'm also affected by chronic laziness!

lisachenko:
> That means that you are betting on Ocramius's good health and well being? He is old, and I heard that he drinks a lot... :D :D :D

> Don't forget that I'm also affected by chronic laziness!

Ah, it's acceptable for me, we can proceed with that, because we are lazy too! :arrow_up:  :smile: :beers:

PR #397, opened 2018-06-07
fix error with "self" references in top level ns

daniel-sc:
This fixes an error when a class with top level namespace uses `self::` to reference a class constant.

The error manifests for example as: 
```
PHP Parse error:  syntax error, unexpected '\' (T_NS_SEPARATOR), expecting identifier (T_STRING)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ composer.lock
 phpunit.xml
 tests/Fixtures/project/var/cache/*
 build/
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ composer.lock
 phpunit.xml
 tests/Fixtures/project/var/cache/*
 build/
+.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ composer.lock
 phpunit.xml
 tests/Fixtures/project/var/cache/*
 build/
+.idea
```
TheCelavi: This should go to your global gitignore. 

See, per example, this: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c but you can find a loot of resources on this topic.

Point here is - rules for ignoring IDE specific files (or system specific files) should not be part of .gitignore files, as we could get into a really large .gitignore files.

TheCelavi (changes requested):
Please revert .gitignore file.

daniel-sc:
Hi @TheCelavi,
I reverted the .gitignore, but for unclear reasons the AppVeyor build fails no (was working before) - could you maybe restart the build?

lisachenko:
AppVeyor is buggy and crappy, so we could ignore this fail.

lisachenko:
I think it would be nice to target 2.x branch, because master is not ready yet, so small patch could be released from 2.x

daniel-sc:
@lisachenko `2.x` and a new release would be perfect!
Can you merge, or do you need any thing else?

Issue #364, opened 2017-09-07
[Weaving] Framework doesn't support weaving of multiple classes in one file


lisachenko:
I have researched fail in `WeavingTransformer` test https://travis-ci.org/goaop/framework/jobs/272452086 and discovered that framework's test weren't updated to configure cache directory which is required now. 

So, when I removed a check for empty cache directory from `WeavingTransformer` and have switched it to VFS, I have discovered serious bug.

Bug is present in calculation of file name for each separate proxy:

```php
$fileName = str_replace(
    $this->options['appDir'] . DIRECTORY_SEPARATOR, 
    '', 
    $class->getFileName()
);
$proxyFileName = $cacheDir . $fileName;
```

Pay an attention to the `$class->getFileName()` which is used for calculation of proxy file name. This expression is based on assumption that each file contains only one class. So, if there are multiple classes in one file, or even worse several namespaces with classes in one single file, then only one last proxy will be saved, all previous proxies will be overridden.

Earlier this feature was working only when cache weren't enabled for framework. Ideally we should drop support of multiple classes/namespaces per one file and only work with PSR-0/PSR-4 standards and Composer. If we can go this way, it can break current users of GoAOP, especially for Yii framework.

TheCelavi:
OH! I see it, proxy is being created on class level, not on file level. 

> drop support of multiple classes/namespaces per one file and only work with PSR-0/PSR-4 

That is the easiest solution, yes.

Otherwise, weaving transformer ought to be re-written in order to support class weaving on file level. Which means pain. You decide.

lisachenko:
@TheCelavi What is your vote for that? Should we keep this for 2.x and fix this issue by adjusting logic of creating proxy files. Or just make it working with single class per file.

Additional fact: framework is fully integrated with composer and operate only on class-level basis, not on file-based. And composer requires PSR-0/PSR-4 standard to be applied to source files where each class MUST be in a separate file without side effects.

TheCelavi:
I vote for "Known issue" explained in documentation. Affected people will complain for sure. If there is noticeable amount of requests to support that, or, if there is a spare time and will to support it, we can write it, why not.

For now, personally, I don't see a reason for support of it, I don't need it.

lisachenko:
Was fixed in 8b2aa19edb518b1d5437765e2c8d34822dfa7c5c

PR #398, opened 2018-07-12
Reflection getFileName fix

andy-shea:
After adding support for multiple classes in a single file (#364), calls to `ReflectionClass->getFilename` within a proxy will return an incorrect path that includes the new namespace fragment.

I've added a test to highlight the issue and included a fix.

The test may need refactoring as the setup (weaving) now occurs in the current thread instead of via a new process.  I did this as I needed access to an instance of the proxy to perform the filename check.  As you can see in the `tearDown` method, I've circumvented some of the side effects of not isolating the proxy creations by resetting the `FilterInjectorTransformer`.  This fixed the only errors that occurred in other tests but I guess it doesn't guarantee that there isn't, or won't be, subtle bugs preventing accurate tests in the future.  Happy to have input here on the best way to achieve this test while still isolating proxy creation.

### DIFF
--- a/src/Instrument/Transformer/MagicConstantTransformer.php
+++ b/src/Instrument/Transformer/MagicConstantTransformer.php
@@ -71,11 +71,14 @@ class MagicConstantTransformer extends BaseSourceTransformer
      */
     public static function resolveFileName(string $fileName): string
     {
-        return str_replace(
+        $suffix = '.php';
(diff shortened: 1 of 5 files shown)


lisachenko (approved):

lisachenko:
Merged, thanks!

PR #406, opened 2019-01-11
Bump phpunit to latest version

icanhazstring:
Also change the test implementation to use new class structure of phpunit.
Remove old configuration settings which are no longer supported.


lisachenko (changes requested):
Nice PR! Thanks!

By the way, could you move FQN `\PhpUnit\Framework\TestCase` to use section to have clean `extends TestCase` string in test classes definitions?

icanhazstring:
Will do

lisachenko (approved):
Great job!

Issue #411, opened 2019-02-08
Allow private property access


jakzal:
It's a slight inconvenience that an AOP framework imposes an implementation on me by forcing my properties to be at least protected.

I'd like to initialise values of annotated properties:

```php
class FindByAspect implements Aspect
{
    /**
     * @Before("@access(Zalas\Annotation\FindBy)")
     */
    public function beforeFieldAccess(FieldAccess $fieldAccess): void
    {
        if (null === $fieldAccess->getValue() && FieldAccess::READ === $fieldAccess->getAccessType()) {
            $field = $fieldAccess->getField()->name;
            $setter = \Closure::bind(function ($object, $value) use ($field) {
                $object->$field = $value;
            }, null, $fieldAccess->getThis());
            $setter($fieldAccess->getThis(), $this->getValue($fieldAccess));
        }
    }

    // ...
}
```

```php
class Foo
{
    /**
     * @FindBy(xpath="//div")
     */
    private $bar;

    public function getBar(): string
    {
        return $this->bar;
    }
}
```

The above code works only if the property is protected or public.

TheCelavi:
Hi, it is more complex than that, actually. GoAOP uses proxy classes to do its weaving in runtime, and proxying is done via inheritance. 

If you are familiar with Java and Spring AOP - similar limitations apply (although, in Spring AOP, proxies are implemented as wrappers, not via inheritance - due to language support). 

When weaving is done in such matter that no proxy is used (original class is modified/weaved) - then access/weaving of private members is possible (like AspectJ). 

Alexander implemented proxy weaving because one of the important feature of any AOP framework must be usage of debugger, and that  is possible only if proxies are used, or some kind of source map is used when original class is compiled/weaved (like in AspectJ, but, AFAIK - no debugging FW for PHP has support for source map).  

I can provide you with more information and/or pointers.

However, I have to close this one, since, this is known limitation of GoAOP and proxy weaving approach in general.

jakzal:
You closed it a bit prematurely perhaps. I created this ticket on @lisachenko's request: http://disq.us/p/1zkpwg8

It is possible to [update the parent's private property](https://3v4l.org/pOkj4) in the proxy class:

```php
<?php

class Foo
{
    private $bar;
    
    public function getBar()
    {
        return $this->bar;
    }
}

class FooProxy extends Foo
{
    public function getBar()
    {
        $this->init();
        return parent::getBar();
    }
    
    private function init()
    {
[...]

PR #412, opened 2019-02-08
[Feature] Add private properties interception

lisachenko:
This PR introduces an ability to intercept access to private properties in classes. We use closure binding in generated proxy child class and bind it to the parent scope so we gain a control over private properties of the class as well.

Here is an example of generated constructor for the `Demo\Example\PropertyDemo` proxy class:
```php
    public function __construct()
    {
        $accessor = function(array &$propertyStorage, object $target) {
            $propertyStorage = [
                'publicProperty' => &$target->publicProperty,
                'protectedProperty' => &$target->protectedProperty,
                'privateProperty' => &$target->privateProperty,
                'indirectModificationCheck' => &$target->indirectModificationCheck
            ];
            unset(
                $target->publicProperty,
                $target->protectedProperty,
                $target->privateProperty,
                $target->indirectModificationCheck
            );
        };
        ($accessor->bindTo($this, parent::class))($this->__properties, $this);
        parent::__construct();
    }
```
As we store references to properties from given original object instance, we can read and write them via aspects without additional code logic.

Feature was requested in #411 by @jakzal

### DIFF
--- a/src/Aop/Framework/ClassFieldAccess.php
+++ b/src/Aop/Framework/ClassFieldAccess.php
@@ -117,14 +117,20 @@ class ClassFieldAccess extends AbstractJoinpoint implements FieldAccess
      */
     public function ensureScopeRule(int $stackLevel = 2): void
     {
-        $property = $this->reflectionProperty;
-
-        if ($property->isProtected()) {
+        $property    = $this->reflectionProperty;
+        $isProtected = $property->isProtected();
+        $isPrivate   = $property->isPrivate();
+        if ($isProtected || $isPrivate) {
             $backTrace     = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, $stackLevel+1);
             $accessor      = $backTrace[$stackLevel] ?? [];
             $propertyClass = $property->class;
             if (isset($accessor['class'])) {
-                if ($accessor['class'] === $propertyClass || is_subclass_of($accessor['class'], $propertyClass)) {
+                // For private and protected properties its ok to access from the same class
+                if ($accessor['class'] === $propertyClass) {
+                    return;
+                }
+                // For protected properties its ok to access from any subclass
+                if ($isProtected && is_subclass_of($accessor['class'], $propertyClass)) {
                     return;
                 }
             }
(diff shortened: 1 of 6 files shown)


src/Aop/Framework/ClassFieldAccess.php
```diff
@@ -117,14 +117,20 @@ public function &getValueToSet()
...
+        $property    = $this->reflectionProperty;
+        $isProtected = $property->isProtected();
+        $isPrivate   = $property->isPrivate();
+        if ($isProtected || $isPrivate) {
             $backTrace     = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, $stackLevel+1);
             $accessor      = $backTrace[$stackLevel] ?? [];
             $propertyClass = $property->class;
             if (isset($accessor['class'])) {
-                if ($accessor['class'] === $propertyClass || is_subclass_of($accessor['class'], $propertyClass)) {
+                // For private and protected properties its ok to access from the same class
```
TheCelavi, line 128: 👍

src/Aop/Framework/ClassFieldAccess.php
```diff
@@ -117,14 +117,20 @@ public function &getValueToSet()
...
+                if ($accessor['class'] === $propertyClass) {
+                    return;
+                }
+                // For protected properties its ok to access from any subclass
+                if ($isProtected && is_subclass_of($accessor['class'], $propertyClass)) {
                     return;
                 }
```
jakzal, line 135: The exception message in line 137 below is probably no longer accurate: `Cannot access protected property...`.

jakzal:
This was quick! Thank you for your hard work! 🍺 

With this PR, my code to set private's property value from an aspect works now:

```php
            $setter = \Closure::bind(function ($object, $value) use ($field) {
                $object->$field = $value;
            }, null, \get_parent_class($fieldAccess->getThis()));
            $setter($fieldAccess->getThis(), 'some value');
```

Not sure if you expect the below code to work as well ('cause it doesn't):

```php
            $fieldAccess->getField()->setAccessible(true);
            $fieldAccess->getField()->setValue($fieldAccess->getThis(), 'some value');

```

src/Aop/Framework/ClassFieldAccess.php
lisachenko: Thanks, somehow have forgotten about this line )

jakzal:
@lisachenko any plans to release this feature? :)

lisachenko:
3.0.0-RC1 is ready to test, if it is ok, then 3.0.0 will be released soon

PR #417, opened 2019-04-09
AA-0003 fixed the Call to undefined method PhpParser\Node\Expr\Variab…

ahmed-abid:
got this issue :
[Error] Call to undefined method PhpParser\Node\Expr\Variable::toString()

while running some tests using AspectMock caused by this line:

https://github.com/goaop/framework/blob/72521d026692131f4052c1d7b3dad15c846cdd0f/src/Instrument/Transformer/MagicConstantTransformer.php#L94

a suggested solution here: https://github.com/nikic/PHP-Parser/pull/596/files

we can do something here as well as following

### DIFF
--- a/src/Instrument/Transformer/MagicConstantTransformer.php
+++ b/src/Instrument/Transformer/MagicConstantTransformer.php
@@ -91,7 +91,7 @@ class MagicConstantTransformer extends BaseSourceTransformer
         /** @var MethodCall[] $methodCalls */
         $methodCalls = $methodCallFinder->getFoundNodes();
         foreach ($methodCalls as $methodCallNode) {
-            if ($methodCallNode->name->toString() !== 'getFileName') {
+            if ($methodCallNode->name instanceof \PhpParser\Node\Identifier && $methodCallNode->name->toString() !== 'getFileName') {
                 continue;
             }
             $startPosition    = $methodCallNode->getAttribute('startTokenPos');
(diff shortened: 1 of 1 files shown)


src/Instrument/Transformer/MagicConstantTransformer.php
```diff
@@ -91,7 +91,7 @@ private function wrapReflectionGetFileName(StreamMetaData $metadata): void
         /** @var MethodCall[] $methodCalls */
         $methodCalls = $methodCallFinder->getFoundNodes();
         foreach ($methodCalls as $methodCallNode) {
-            if ($methodCallNode->name->toString() !== 'getFileName') {
+            if ($methodCallNode->name instanceof \PhpParser\Node\Identifier && $methodCallNode->name->toString() !== 'getFileName') {
```
lisachenko: Please use import of class name

lisachenko (changes requested):
Hi, thank you fo sending this PR, please fix small notice in it.

src/Instrument/Transformer/MagicConstantTransformer.php
ahmed-abid: Please review

lisachenko (approved):

lisachenko:
Squashed and merged into master, thank you!

ddinchev:
@lisachenko I'm also hitting the same issue. Can you release this fix in 2.3.x?

lisachenko:
@ddinchev This fix was merged into the master branch only, so you should not expect it to be in 2.x branch. Please check **master** branch instead, aka **3.x-dev**

PR #421, opened 2019-06-26
Finder glob pattern

Rastusik:
The former change of spl file iterators to symfony finder didn't have support for glob patterns (or possibly broke the former support) 

This change is supposed to add support for glob patterns into the include_paths config section

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ composer.lock
 phpunit.xml
 tests/Fixtures/project/var/cache/*
 build/
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ composer.lock
 phpunit.xml
 tests/Fixtures/project/var/cache/*
 build/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ composer.lock
 phpunit.xml
 tests/Fixtures/project/var/cache/*
 build/
+.idea
```
lisachenko: Please, don't add IDE-specific folders into the project's .gitignore. Configure global gitignore instead, see https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

lisachenko (changes requested):

.gitignore
Rastusik: fixed

lisachenko (approved):

lisachenko:
Squashed and merged into 2.x, thanks!

PR #422, opened 2019-07-30
Memory leaks / method invocation side effects.

badrutdinovrr:
First of all, thank you for your awesome framework!
Unfortunately, a method \Go\Aop\Framework\AbstractMethodInvocation::__invoke doesn't free references to method arguments as soon as have been invoked which results in:
1. Memory leaks
2. Unintended side effect: the argument object destructor isn't called.

I hope my small pull request will help your cool project to become a bit better.
Thank you!

### DIFF
--- a/src/Aop/Framework/AbstractMethodInvocation.php
+++ b/src/Aop/Framework/AbstractMethodInvocation.php
@@ -92,6 +92,9 @@ abstract class AbstractMethodInvocation extends AbstractInvocation implements Me
 
             if ($this->level > 0) {
                 [$this->arguments, $this->instance, $this->current] = array_pop($this->stackFrames);
+            } else {
+                $this->arguments = [];
+                $this->instance  = null;
             }
         }
     }


lisachenko:
Hello! Thank you for the PR.

As master branch is not ready for release yet, please rebase your PR on 2.x branch, so I will be able to release a quick patch.

Also, there are some issues in review, please fix them before merging

src/Aop/Framework/AbstractMethodInvocation.php
```diff
@@ -93,6 +93,11 @@ public function __construct(string $className, string $methodName, array $advice
             if ($this->level > 0) {
                 [$this->arguments, $this->instance, $this->current] = array_pop($this->stackFrames);
             }
+            else {
```
lisachenko: Put **else** on the previous line after figure bracket.

src/Aop/Framework/AbstractMethodInvocation.php
```diff
@@ -93,6 +93,11 @@ public function __construct(string $className, string $methodName, array $advice
...
+                $this->arguments = [];
+                $this->instance = null;
+                $this->current = null;
```
lisachenko: Please, align all consecutive assignments. Also, `$this->current `is updated each time in the recursive calls, I'm not sure that your patch support recursive calls. I would suggest to drop this line entirely, clear only arguments and instance variables

lisachenko (changes requested):

badrutdinovrr:
Closing this in favor of #423

PR #427, opened 2019-08-11
[Bug-425] Special treatment for initialization advices in order to fi…

l00k:
…x wrong class construction issue

### DIFF
--- a/src/Proxy/ClassProxyGenerator.php
+++ b/src/Proxy/ClassProxyGenerator.php
@@ -134,7 +134,7 @@ class ClassProxyGenerator
 
         $joinPointsProperty->setAccessible(true);
         $advices    = $joinPointsProperty->getValue();
-        $joinPoints = static::wrapWithJoinPoints($advices, $reflectionClass->getParentClass()->name);
+        $joinPoints = static::wrapWithJoinPoints($advices, $reflectionClass->getParentClass()->name, $targetClassName);
         $joinPointsProperty->setValue($joinPoints);
 
         $staticInit = AspectContainer::STATIC_INIT_PREFIX . ':root';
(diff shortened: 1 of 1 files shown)


lisachenko:
Hi! PR is being verified and reviewed

src/Proxy/ClassProxyGenerator.php
```diff
@@ -177,6 +177,23 @@ protected static function wrapWithJoinPoints(array $classAdvices, string $classN
         }
 
         $joinPoints = [];
+        
+        // special treatment for init advices
```
lisachenko, line 181: I don't like this extra copy-paste logic only for one type of advices. Should be rewritten in another way. This solution is not good.

src/Proxy/ClassProxyGenerator.php
```diff
@@ -166,7 +166,7 @@ public function generate(): string
      *
      * @return Joinpoint[] returns list of joinpoint ready to use
      */
-    protected static function wrapWithJoinPoints(array $classAdvices, string $className): array
+    protected static function wrapWithJoinPoints(array $classAdvices, string $className, string $constructorClassName): array
```
lisachenko, line 169: If we follow this way, we It would be nice to give better naming of variables: `$parentClassName` and `$originalClassName` but I think we can leave this code as-is and modify only the `ReflectionConstructorInvocation` to know which class is used now.

lisachenko (changes requested):
I can consider this PR as a quick-fix, but it can't be merged due to the explicit copy-paste logic for one single case.

We can continue in two ways:
1) you can research how to do this in the `ReflectionConstructorInvocation` + minimum changes in other files to integrate this and update this PR for merge
2) I can write a fix myself and merge it, closing this PR

lisachenko:
Also, need to target 2.x branch to release a quick-fix here as well (while waiting for 3.0.0 release)

l00k:
Yea.. I know that it was not nice solution. Please fix it yourself :)

lisachenko:
> Yea.. I know that it was not nice solution. Please fix it yourself :)

Close this one in favor of #428.

Anyway, thank you for reporting and looking into this, because it's rarely used feature.
PS. Previously I had a wish to remove this feature it in the next major )

l00k:
Ouu no ;) it is useable for DI implementation. It could be done without that, but it is a lot easier with constructor modification.

PR #440, opened 2020-01-25
Allow installation of symfony/finder 5.x component 

TonySchmitt:
I have run the phpunit and I don't have error with the support of symfony 5.0. So, I think we can add this support.


lisachenko:
Ok, thank you, merged into master, aka 3.x-dev

PR #444, opened 2020-02-29
Use laminas-code instead of zend-code

Naktibalda:
Zend Framework was rebranded to Laminas last year and all Zend libraries got new names.

Now `composer install` prints these warnings:

> Package zendframework/zend-code is abandoned, you should avoid using it. Use laminas/laminas-code instead.
> Package zendframework/zend-eventmanager is abandoned, you should avoid using it. Use laminas/laminas-eventmanager instead.


lisachenko (approved):

lisachenko:
Thank you for the PR

PR #447, opened 2020-04-03
[Compatibility] Allow Symfony 5 components.

func0der:


func0der:
This change is for a 2.x release, since 3.x is yet to come, a quick 2.x would be nice :)

func0der:
@lisachenko Do you have an idea how to fix that build check?

lisachenko:
> Do you have an idea how to fix that build check?

Hmm, this can be an issue in the SymfonyInsight itself, as it tries to install version of package with incorrect PHP version.

lisachenko:
Also, I don't see build results from Travis...

lisachenko:
Tagged patch release 2.3.4

PR #473, opened 2021-04-23
old bug sync

137-rick:
allow the error type with throwable exception


137-rick:
ops.... how to fix that error

lisachenko:
Merged into 2.x. Master has already contained proper type.

PR #483, opened 2022-04-19
[FIX] Library generates non-nullable return type for nullable return …

jakublabno:
…methods

Library version 3.0.0

For below method proxy is generating non-null return
```
/**
     * @AopAnnotation()
     * @return null|int
     */
    public function getSomeRandomValueAllowOnNull(): ?int
    {
        return random_int(0, 666);
    }
```
result:
```
/**
     * @AopAnnotation()
     * @return null|int
     */
    public function getSomeRandomValueAllowOnNull() : int
    {
        return self::$__joinPoints['method:getSomeRandomValueAllowOnNull']->__invoke($this);
    }
```


jakublabno (commented):
t

PR #486, opened 2022-11-23
PHP 8 Support

WalterWoshid:
I have added PHP 8 Support by replacing the `@Annotation` with the new `#[Attribute]` syntax

I have refactored `goaop/parser-reflection` to also support PHP 8. I might have refactored too much of it, but all the tests are running :)

The tests from the framework must be edited, but most of them work as well. I also ran some custom tests on the Laravel framework and it seemed to work

Tasks:
 - [x] Publish new version of `jakubledl/dissect` and register it in the composer - current version is a little bit outdated, but still works.
 - [ ] Implement in the `goaop/parser-reflection` support for PHP8. Mostly for existing functional, need to add support for parsing attributes from the AST data and try not to call original reflection instances to avoid autoloading. Support for PHP8 enums is missing, but I guess this can be added later separately. This new version should be published as a new major with bump of minimum PHP version to >=8.0
 - [ ] Bump dependencies here, in the `goaop/framework` to include new version of `goaop/parser-reflection`, `dissect` libraries. Also, bump PHP minimum version to use PHP>=8.0.
 - [ ] Remove doctrine annotations reader and related parts, as attributes are finally in the core of PHP. There are some parts with caching of annotations that should be removed too.


demos/Demo/Example/ErrorDemo.php
```diff
@@ -4,15 +4,15 @@
...
+ * @copyright Copyright 2014-2022, Lisachenko Alexander <lisachenko.it@gmail.com>
  *
  * This source file is subject to the license that is bundled
  * with this source code in the file LICENSE.
  */
 
 namespace Demo\Example;
 
-use Demo\Annotation\Deprecated as deprecated; // <== We import specific system annotation
+use Demo\Attribute\Deprecated as deprecated; // <== We import specific system annotation
```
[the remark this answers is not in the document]
WalterWoshid: There is a blank line after the last USE statement

WalterWoshid:
There are still some things to do after this:

- [x] Replace `jakubledl/dissect` with `walterwoshid/dissect` which is a fork that supports PHP 8
- [ ] Add `__unserialize` and `__serialize` to remove deprecation warning in console for PHP 8.1
- [ ] `ReflectionClassLikeTrait::newInstance` needs different types for PHP 8.0 and PHP 8.1 - how would one implement that?

src/Core/GeneralAspectLoaderExtension.php
```diff
@@ -53,15 +52,19 @@ public function load(Aspect $aspect, ReflectionClass $reflectionAspect): array
         $loadedItems = [];
         foreach ($reflectionAspect->getMethods() as $aspectMethod) {
             $methodId    = $reflectionAspect->getName() . '->'. $aspectMethod->getName();
-            $annotations = $this->reader->getMethodAnnotations($aspectMethod);
+            $annotations = $aspectMethod->getAttributes(
```
lisachenko: I guess, that `reader` property now can be removed and entire doctrine annotation reader too from the list of dependencies.

lisachenko: `AbstractAspectLoaderExtension` can be patched too to remove reader

lisachenko: Have you also implemented support for reading attributes in `goaop/parser-reflection`, as this part of code should work without touching real reflection, thus entire attribute support is required to be implemented via AST analysis only.

lisachenko (changes requested):

lisachenko:
Helo, @WalterWoshid! Really nice work towards PHP8 support, check my small comments inside PR.

composer.json
```diff
@@ -4,17 +4,17 @@
...
-    "homepage": "http://go.aopphp.com/",
+    "homepage": "https://go.aopphp.com/",
     "license": "MIT",
 
     "require": {
-        "php": "^7.4.0",
+        "php": "^8.0",
         "doctrine/annotations": "^1.11.1",
         "doctrine/cache": "^1.10",
         "goaop/parser-reflection": "^3.0.1",
```
lisachenko, line 14: I don't see any support for PHP8 attributes here inside `goaop/parser-reflection` and this was main blocker for me to get PHP8 support, have you implemented support for PHP8 inside `goaop/parser-reflection`?

src/Core/GeneralAspectLoaderExtension.php
WalterWoshid: You're right, some things can be removed now, I only aimed for a functioning base for the framework to support PHP 8

WalterWoshid: I did some changes to the `goaop/parser-reflection`, but I am not really sure what I did there, all tests were passing though

composer.json
WalterWoshid: I tried to remove the `goaop/parser-reflection` as you can see in my [fork-2 branch](https://github.com/WalterWoshid/goaop-framework/tree/fork-2), but somehow was not able to fully understand Better-Reflection to resolve parent classes (extends Class), because it always threw an Exception, that it couldn't find it in the [SourceLocator](https://github.com/Roave/BetterReflection/blob/6.5.x/docs/usage.md#sourcelocators)

See [getParent](https://github.com/WalterWoshid/goaop-framework/blob/7b3fe734a07f2f73177c451cc60a04213559296a/src/Core/Service/AdviceMatcher.php#L98) and [$parent->getProperties(...)](https://github.com/WalterWoshid/goaop-framework/blob/7b3fe734a07f2f73177c451cc60a04213559296a/src/Core/Service/AdviceMatcher.php#L170)

So I only upgraded the parameter and return types to support PHP 8 and PHP 8.1

WalterWoshid: And those changes don't have backwards compatibility for older versions. I think supporting the [currently supported PHP version](https://www.php.net/supported-versions.php) is more important than providing support for all versions, which in my opinion I would want to, but it's too difficult to focus on multiple versions

WalterWoshid:
I would also like to hear your thoughts about a new and easier syntax style proposal:

```php
<?php

use Go\Aop\Intercept\MethodInvocation;
use Go\Lang\Attribute\Aspect;
use Go\Lang\Attribute\Before;

#[Aspect]
class AspectExample /* or keep the old */ implements Aspect /* instead of the #[Aspect] */
{
    #[Before(
        public: true,
        class: TargetClass::class, // or "Namespace/Target*"
        method: "render", // or "get*"
    )]
    public function beforeMethodExecution(...)
    {
        // ...
    }
}
```

It would use the new [named arguments syntax for PHP 8](https://stitcher.io/blog/php-8-named-arguments) which adds better typings and you could skip the whole `execution(...)` parsing, which would be faster I guess. It requires a lot of refactoring though

composer.json
lisachenko: Depends on https://github.com/goaop/parser-reflection/pull/122

lisachenko: I think we can create a new major versions based on your work in both projects. Then we will have an opportunity to add `branch-alias` and RC/beta tags to test them together and put a requirement on both of them to work with PHP>=8.0. Keeping BC with version 7.4 is not required at all.

lisachenko:
> I would also like to hear your thoughts about a new and easier syntax style proposal:
> 
> ```
> <?php
> 
> use Go\Aop\Intercept\MethodInvocation;
> use Go\Lang\Attribute\Aspect;
> use Go\Lang\Attribute\Before;
> 
> #[Aspect]
> class AspectExample /* or keep the old */ implements Aspect /* instead of the #[Aspect] */
> {
>     #[Before(
>         public: true,
>         class: TargetClass::class, // or "Namespace/Target*"
>         method: "render", // or "get*"
>     )]
>     public function beforeMethodExecution(...)
>     {
>         // ...
>     }
> }
> ```
> 
> It would use the new [named arguments syntax for PHP 8](https://stitcher.io/blog/php-8-named-arguments) which adds better typings and you could skip the whole `execution(...)` parsing, which would be faster I guess. It requires a lot of refactoring though

It looks good for me, assuming native attributes support - this looks logical 👍

WalterWoshid:
@lisachenko I read that you don't have much time to manage this framework. If you'd like, you could give me concrete tasks to do and so I could do them for you )

lisachenko:
Oh, that's sounds sweet for me ) So, basically, you have almost done full the job already, let's collect and put all tasks one by one. Writing tasks in your main PR body now...

AmraniCh:
hey, any news for PHP 8 support ?

AmraniCh:
@WalterWoshid I already did, I think I do not have other options because our web application is running on PHP 8, however I would like to thank you for your effort in this package 👍 
If I see something needs to improve or fix I will open a PR!

HighLiuk:
> Hi @AmraniCh, check out `Okapi/AOP` which is my own implementation of AOP for PHP :)
> 
> https://github.com/okapi-web/php-aop

@WalterWoshid You saved my life! I starred your repo like number 1 🔝

PR #490, opened 2023-05-25
fix: prevent generating "namespace ;" for classes without declared namespace

daniel-sc:
This fixes a bug when trying to work with classes that have no namespace declared - see updated unit test.

Note that the error comes as an rather unspecific
```
ParseError: syntax error, unexpected ';', expecting '{'
```
As the generated code includes `namespace ;`!

It would be very nice if this could be released as a bugfix soon :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,8 @@
             "Go\\Tests\\TestProject\\": "tests/Fixtures/project/src/"
         },
         "files": [
-            "tests/functions.php"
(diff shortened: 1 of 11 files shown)


daniel-sc:
@lisachenko Could you look into this PR? I'd be happy to address any issues!

lisachenko:
Thanks! Finally, merged ⏳

PR #502, opened 2024-04-14
[PHP8] Add broken test with failing attributes generation

lisachenko:
This broken test is used to track issue with broken codogen with attributes near class/method/method/function/property/parameter.


lisachenko:
Blocked on https://github.com/laminas/laminas-code/pull/145

lisachenko:
FYI @samsonasik this issues is considered as main blocker for running on codebase with attributes - child class won't contain any attributes and this breaks the logic of clients who expect attributes to be preserved.

lisachenko:
One solution can be switch to something under our control (I doubt that I'll have enough resources to support this).

Another solution is to completely remove dependency on laminas package and instead use PrettyPrinter and AST nodes from the `nikic/php-parser` via `BuilderFactory::attribute` and similar to generate new code.

samsonasik:
@lisachenko I am currently on my village and will be back on 16th/17th April, so hopefully can fully looking at it after that day as possibly 2 days long trip.

Ceate own builder for it can be solution, but need to check how many laminas-code already involved in other area

samsonasik:
@lisachenko it seems too tight with laminas-code, and for the alternative, to be able to use php-parser's Printer, we need to know the Node from the `ReflectionFunction` which may need to use `Go\ParserReflection\ReflectionFunction`, since `Function_` parameter is required, we may scan from very early with `Go\ParserReflection\ReflectionFileNamespace`, I don't have idea how big the refactor it can be, any idea?

lisachenko:
Fixed in the master branch

PR #503, opened 2024-04-14
[PHP8] Add broken tests with complex types

lisachenko:
This functional test highlights some internal issues inside `SelfValueVisitor` when dealing with complex types (union/intersection/DNF).

Framework should be able to work with any complex types transparently.

PR #504, opened 2024-05-05
Add PHP 8.0+ Union and Intersection type support on SelfValueVisitor

samsonasik:
Closes https://github.com/goaop/framework/pull/503


tests/Go/Instrument/Transformer/_files/php82-file-transformed.php
```diff
@@ -0,0 +1,94 @@
...
+ * @see https://php.watch/versions/8.2/constants-in-traits
+ */
+trait TraitWithPhp82Constant
+{
+    protected const CURRENT_VERSION = '2.6';
+    final protected const MIN_VERSION = '2.5';
+
+    protected function ensureVersion(): void
+    {
+        if (\Go\ParserReflection\Stub\ClassWithPhp82NullFalseTypes::CURRENT_VERSION < \Go\ParserReflection\Stub\ClassWithPhp82NullFalseTypes::MIN_VERSION) {
```
samsonasik: on trait, the transformation seems incorrect, even on public constant, see 

- https://3v4l.org/uVC61#v8.2.18
- https://3v4l.org/YWIMP#v8.2.18

keep self seems the way to go: https://3v4l.org/aS44R#v8.2.18

samsonasik:
@lisachenko could you review this? Thank you.

tests/Go/Instrument/Transformer/_files/php82-file-transformed.php
lisachenko: I think, that this behaviour deserves separate issue to track, otherwise it can break code easily, I would recommend now to adjust logic of the `SelfValueVisitor` to skip code transformation if `self` is inside trait and not in the class.

lisachenko:
PS. The more I look on the `SelfValueVisitor`/`SelfValueTransformer` classes - the more I think that they do very strange and incorrect job, which I would like to completely rewrite to enable pretty-printing and to switch to AST-only transformation. 

FYI: I'm now in the process of slow refactoring of entire AST/transformation subsystem, so expect significant changes in this part soon )

Could you please check if current PR works with the latest master branch or not (try to rebase and check locally please)? I've some bad feeling that there is a regression issue from my recent refactoring of `goaop/parser-reflection` that renders 'self' now with '\self' (

samsonasik:
Using dev-master seems working ok

Screenshot 2024-05-06 at 05 41 33

tests/Go/Instrument/Transformer/_files/php82-file-transformed.php
samsonasik: I tried to use `ParentConnectingVisitor` on  `SelfValueTransformer` to allow traverse up to verify it inside Trait, but it somehow can't get parent on target `self`, do you have idea to verify it ?

lisachenko: I would better go with extra variable/check ` elseif ($node instanceof Trait_) {` and store this information as boolean flag `isInsideTrait` to stop transformation inside traits at all.

samsonasik: sure, I updated to verify inside trait or enum, as the bug is happen on enum and trait https://github.com/goaop/framework/pull/504/commits/e13648e1bb9522c9911b56ec26a38f26259153df

samsonasik:
@lisachenko Ready to merge 👍

tests/Go/Instrument/Transformer/_files/php82-file-transformed.php
samsonasik: Ok, I found that the bug is due to `className` property re-use when single file consist of multiple `ClassLike`, I create a real patch for it https://github.com/goaop/framework/pull/504/commits/19c8ce81f13084f7be101d4a62daccaccf5574aa to make it null on `Trait_`

samsonasik:
@lisachenko Ready to merge 👍

samsonasik:
Also handled `self` usage in anonymous class

samsonasik:
@lisachenko Ready to merge 👍

lisachenko (approved):

lisachenko:
Squashed and merged into the main branch. Thanks!

PR #505, opened 2024-05-19
Added DNF type support

krzysztof-ciszewski:
TODO:
* tests
* domain exceptions


krzysztof-ciszewski:
@lisachenko I would suggest disabling nitpick, seems pretty outdated, most of these comments are wrong

lisachenko:
Outdated, not compatible with new core. Also, I'm not sure that AOP expressions need to be complex - this will create very hard-to-debug expressions and code

PR #541, opened 2026-04-18
feat: Create copilot-setup-steps.yml

lisachenko:
Adds a dedicated GitHub Actions workflow to provision PHP 8.4 + Composer dependencies for the Copilot agent so repository tests can run successfully under the required PHP version.

**Changes:**
- Introduces `.github/workflows/copilot-setup-steps.yml` with `copilot-setup-steps` job.
- Checks out the repo, installs PHP 8.4 via `shivammathur/setup-php`, and runs `composer install`.

More details is here: https://docs.github.com/en/enterprise-cloud@latest/copilot/how-tos/use-copilot-agents/cloud-agent/customize-the-agent-environment

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,44 @@
+name: "Copilot Setup Steps"
+
+# Automatically run the setup steps when they are changed to allow for easy validation, and
+# allow manual testing through the repository's "Actions" tab
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+
+jobs:
+  # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
+  copilot-setup-steps:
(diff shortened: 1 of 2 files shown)


lisachenko:
@copilot run tests and report in comment status of tests, then run phpstan too as well

lisachenko:
@copilot, setup GitHub Composer Authentication

By default, setup-php uses the GITHUB_TOKEN secret that is generated for each workflow run. In case you want to use a Personal Access Token (PAT) instead, you can set the github-token input.

- name: Setup PHP
  uses: shivammathur/setup-php@v2
  with:
    php-version: '8.4'
    github-token: ${{ secrets.GH_PAT_TOKEN }}
The COMPOSER_TOKEN and GITHUB_TOKEN environment variables have been deprecated in favor of the github-token input and will be removed in the next major version.

Update to use value of secrets.GH_PAT_TOKEN which I've added.

Also, update instruction for agent to use PHP8.4 or higher for agents. If version is PHP8.3 - then stop, as you will not be able to validate result with tests/phpstan.

lisachenko:
Ok, I see - it will work only from the master branch. Once merged - it will be used for subsequent builds.

PR #551, opened 2026-04-24
feat: Add return type as second generic variable to core interfaces

lisachenko:
Extend core interfaces to support generic return type as second template variable.

MethodInvocation uses V as generic return type
FunctionInvocation uses V as generic return type
FieldAccess uses V as generic property type

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -79,13 +79,13 @@ class Foo extends OriginalParent implements OriginalInterfaces, \Go\Aop\Proxy
         // ... one alias per intercepted method (including private ones)
     }
 
-    public function interceptedMethod(...) {
-        /** @var \Go\Aop\Intercept\DynamicMethodInvocation<self>|null $__joinPoint */
+    public function interceptedMethod(ArgType $arg): ReturnType {
+        /** @var \Go\Aop\Intercept\DynamicMethodInvocation<self, ReturnType>|null $__joinPoint */
         static $__joinPoint;
         if ($__joinPoint === null) {
             $__joinPoint = \Go\Aop\Framework\InterceptorInjector::forMethod(self::class, 'interceptedMethod', [...]);
         }
-        return $__joinPoint->__invoke($this, [...]);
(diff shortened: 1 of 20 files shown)


lisachenko:
@copilot please add logic that will generate phpDoc annotation for intercepted methods with second generic variable type if method has any return type specified, including "void":

```php
    protected function perform(string $task, string $level): void
    {
        /** @var \Go\Aop\Intercept\DynamicMethodInvocation<self, void>|null $__joinPoint */
        static $__joinPoint;
        if ($__joinPoint === null) {
            $__joinPoint = \Go\Aop\Framework\InterceptorInjector::forMethod(self::class, 'perform', ['Demo\Aspect\LoggingAspect->beforeMethodExecution']);
        }
        $__joinPoint->__invoke($this, [$task, $level]);
    }

    public function getReport(string $from): string
    {
        /** @var \Go\Aop\Intercept\DynamicMethodInvocation<self, string>|null $__joinPoint */
        static $__joinPoint;
        if ($__joinPoint === null) {
            $__joinPoint = \Go\Aop\Framework\InterceptorInjector::forMethod(self::class, 'getReport', ['Demo\Aspect\CachingAspect->aroundCacheable']);
        }
        return $__joinPoint->__invoke($this, [$from]);
    }
```

Same should be for static methods and for functions: if original item has a return type - it should be added as generic into phpDoc block.

Also, fix phpstan error

lisachenko:
@copilot update also the CLAUDE.md file, include short information about generic templates and how return type is used to ensure type-awareness. Also, update examples (if any)

lisachenko:
@copilot rebase this branch on top of origin/master and resolve the conflicts in this pull request, then force-push rewritten history again to the same branch

lisachenko:
@copilot check and fix unit tests on PHP8.5

lisachenko:
@copilot please do not make any assumptions about the return type itself, just take the existing type definition (ideally from AST node) and put it into the type definition as-is, this will be much easier. Do not patch anything in the vendor libraries, if you want to do these changes - ask permission first

PR #569, opened 2026-05-08
[Feature] support PHP 8.5 first-class callable as default value for parameters and properties

lisachenko:
## Summary
- Proxy generation now preserves FCC default expressions (e.g. `function foo($cb = strlen(...))`) via raw AST nodes
- `ValueGenerator` — added `fromExprNode()` factory for pre-built AST `Expr` nodes that can't be represented as PHP scalars
- `PropertyGenerator` — added `setDefaultExpressionNode()` for `Expr`-based property defaults
- `ParameterGenerator::fromReflection()` — uses `getNode()->default` directly when parser-reflection is available, avoiding parser-reflection bugs where `getDefaultValue()` crashes (uninitialized typed property) or returns `null` for Closure defaults
- `AbstractInterceptedPropertyGenerator::createBasePropertyGenerator()` — same AST-first approach for intercepted properties

## Why AST-first?
When parser-reflection is loaded, the raw AST default node is the authoritative source — it works uniformly for scalars, arrays, and FCC expressions. Calling `getDefaultValue()` can crash parser-reflection (`ReflectionParameter::$defaultValue` is uninitialized for FCC) or silently return `null` (`ReflectionProperty::$defaultValue` defaults to `null`). By reading `getNode()->default` directly, we bypass these issues entirely.

## Test plan
- [x] Full test suite: 2467 tests, 0 failures, 0 errors
- [x] PHPStan level 10: 0 errors
- [x] Array property interception test passes (validates AST path works for non-Closure const-expr defaults)
- [x] PHP 8.4 safety: all new code paths gated by `method_exists($param, 'getNode')` (parser-reflection only)


lisachenko:
Depends on https://github.com/goaop/parser-reflection/pull/193

PR #584, opened 2026-07-29
fix: emit fully qualified names for global-namespace symbols in class proxies

lisachenko:
## Summary

Fixes a class-loading crash when a `DeclareParents` introduction adds a **global-namespace** interface or trait (e.g. `Stringable`) to a namespaced class. `ClassGenerator` emitted single-segment names as *relative* (`implements Stringable`), so inside the proxy's namespace PHP resolved them as `My\Ns\Stringable` and failed with `Interface "My\Ns\Stringable" not found`. Multi-segment names were unaffected, which is why ordinary advices never hit this.

Found while triaging goaop/goaop-laravel-bridge#21 (inter-type declarations failing under Laravel): the legacy 2.x report was a dissect-grammar parse error, but reproducing the scenario on 4.x surfaced this current bug — the end-to-end DeclareParents weaving test in the bridge (goaop/goaop-laravel-bridge#35) fails without this patch and passes with it.

## Changes

- `src/Proxy/Generator/ClassGenerator.php`:
  - Interfaces are now always emitted as `Name\FullyQualified` — they are documented as FQCNs, and this matches the identical handling (and comment) already present in `EnumGenerator`.
  - Parent class and trait names honour an explicit leading backslash (`\Exception`, `\GlobalTrait`) as "fully qualified", while bare short names (`Foo__AopProxied`, same-namespace parents) keep resolving relative to the proxy's namespace, preserving the existing behavior covered by `testExtendsSimpleName`.
- `tests/Proxy/Generator/ClassGeneratorTest.php`: three new tests covering a global-namespace interface in a namespaced proxy, an explicitly-rooted global parent, and an explicitly-rooted global trait mixed with a deliberate short trait name.

## Testing

- `./vendor/bin/phpunit` — 2474 tests, 2877 assertions, green (20 pre-existing deprecations, 1 pre-existing skip)
- `phpstan analyze --memory-limit=512M` — level 10, no errors
- End-to-end: goaop/goaop-laravel-bridge#35 `DeclareParentsWeavingTest` fails without this fix, passes with it

https://claude.ai/code/session_01KtrBW3xd8DQkxmHvX9qr4p

---
_Generated by [Claude Code](https://claude.ai/code/session_01KtrBW3xd8DQkxmHvX9qr4p)_

### DIFF
--- a/src/Proxy/Generator/ClassGenerator.php
+++ b/src/Proxy/Generator/ClassGenerator.php
@@ -154,6 +154,24 @@ final class ClassGenerator implements GeneratorInterface
     }
 
     /**
+     * Builds the AST name node for a class-like reference (parent, interface,
+     * trait, alias) with a single universal rule: explicitly rooted
+     * ("\Stringable") and multi-segment names are fully qualified; a bare
+     * short name resolves in the generated class's own namespace (e.g. the
+     * Foo__AopProxied body trait). Global-namespace names coming from
+     * ::class constants are rooted upstream (AdviceMatcher, proxy generators)
+     * before they reach this generator.
+     */
+    private static function classNameNode(string $name): Name
+    {
+        $normalized = ltrim($name, '\\');
+
+        return str_starts_with($name, '\\') || str_contains($normalized, '\\')
+            ? new Name\FullyQualified($normalized)
+            : new Name($normalized);
+    }
+
+    /**
      * Returns the class AST node only — no namespace or use wrappers.
      * Suitable for direct injection into a cloned file AST.
      */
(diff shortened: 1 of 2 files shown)


src/Proxy/Generator/ClassGenerator.php
```diff
@@ -176,35 +176,39 @@ public function getNode(): ClassNode
...
 
         if ($this->parentClass !== null && $this->parentClass !== '') {
-            // Use FullyQualified when the name contains a namespace separator to avoid
-            // ambiguity in the generated file's namespace context.
+            // Use FullyQualified when the name is explicitly rooted ("\Exception") or
+            // contains a namespace separator; a bare short name stays relative so
+            // same-namespace parents keep resolving in the generated file's context.
             $parentName = ltrim($this->parentClass, '\\');
-            $parentNode = str_contains($parentName, '\\')
+            $parentNode = str_starts_with($this->parentClass, '\\') || str_contains($parentName, '\\')
```
lisachenko: make whole check here around parentName/parentClass consistent - not it looks tricky and unclear. Keep minimum readable version.

src/Proxy/Generator/ClassGenerator.php
```diff
@@ -176,35 +176,39 @@ public function getNode(): ClassNode
...
         }
 
         foreach ($this->interfaces as $interface) {
             if ($interface !== '') {
-                $ifaceName = ltrim($interface, '\\');
-                $ifaceNode = str_contains($ifaceName, '\\')
-                    ? new Name\FullyQualified($ifaceName)
-                    : $ifaceName;
-                $builder->implement($ifaceNode);
+                // Interfaces are FQCNs by contract — always emit fully qualified so a
```
lisachenko: I don't agree with this assumption, logic should be the same for extends/implements/traits adoptation, etc

therefore, need a universal solution that works universally for all class-string names in the file.

Maybe we need somehow patch the DeclareParents instead, assuming that it should use constant-string expressions like `Stringable::class` and all `::class` strings do not use FQN prefix "\" - we might add it here. This will work with imports, relative declare parents, etc

src/Proxy/Generator/ClassGenerator.php
lisachenko: Done in 72b14d3 — the inline parent check is gone; `extend`, `implement`, trait names and trait aliases all go through one small `classNameNode()` helper now, so every site reads the same.

---

src/Proxy/Generator/ClassGenerator.php
lisachenko: Agreed, and done in 72b14d3: extends/implements/trait names/trait aliases now share one universal rule via `classNameNode()` — explicitly rooted (`\Stringable`) or multi-segment names are emitted `FullyQualified`, bare short names resolve in the generated class's own namespace (which keeps the same-namespace `Foo__AopProxied` body trait working).

On patching `DeclareParents` instead: the normalization you describe already exists — `AdviceMatcher::getIntroductionAdvices()` does `'\\' . ltrim($name, '\\')` on both the introduced interface and trait, so every `::class` string (import-based or relative, since `::class` always yields the resolved FQCN without a leading `\`) reaches the generator explicitly rooted. The bug was only that the generator then dropped that marker via `ltrim` before deciding qualification. With the marker honored by the universal rule, both ends line up: attribute-side names are rooted once at load time, and the generator has a single consistent emission rule for every class-string in the file.

CI is green on the refactor (2474 tests + phpstan level 10), and the end-to-end bridge test in goaop/goaop-laravel-bridge#35 still passes against this branch locally.

---

PR #585, opened 2026-07-30
Add PHP 8.6 to CI matrix as experimental

lisachenko:
## Summary

Adds PHP 8.6 to the PHPUnit workflow matrix as an experimental version, so early failures on the upcoming PHP release are visible without breaking the build:

- New `experimental` matrix flag (`false` for the existing PHP 8.4 / 8.5 jobs).
- `include` entries running PHP 8.6 with both `lowest` and `highest` dependencies, marked `experimental: true`.
- `continue-on-error: ${{ matrix.experimental }}` on the job, so PHP 8.6 failures don't fail the workflow.
- `fail-fast: false` so one failing combination no longer cancels the remaining matrix jobs.

The `composer.json` constraint (`"php": "^8.4.0"`) already allows 8.6, so no dependency changes are needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_0196qA77VM2Kxaqu6YoiuMLC)_

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -15,7 +15,10 @@ jobs:
 
     runs-on: ${{ matrix.operating-system }}
 
+    continue-on-error: ${{ matrix.experimental }}
+
     strategy:
+      fail-fast: false
       matrix:
         dependencies:
           - "lowest"
@@ -25,6 +28,17 @@ jobs:
           - "8.5"
         operating-system:
           - "ubuntu-latest"
+        experimental:
+          - false
+        include:
+          - php-version: "8.6"
+            dependencies: "lowest"
+            operating-system: "ubuntu-latest"
+            experimental: true
+          - php-version: "8.6"
+            dependencies: "highest"
+            operating-system: "ubuntu-latest"
+            experimental: true
 
     steps:
       - name: "Checkout"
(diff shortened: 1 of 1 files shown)


.github/workflows/phpunit.yml
```diff
@@ -49,15 +63,15 @@ jobs:
 
       - name: "Install lowest dependencies"
         if: ${{ matrix.dependencies == 'lowest' }}
-        run: "composer update --prefer-lowest --no-interaction --no-progress"
+        run: "composer update --prefer-lowest --no-interaction --no-progress${{ matrix.experimental && ' --ignore-platform-req=php+' || '' }}"
```
lisachenko, line 66: Why do we need to ignore platform requirements?

lisachenko: Without it, the PHP 8.6 jobs fail during `composer update`, before PHPUnit ever runs: `tracy/tracy` (all versions up to v2.12.0) declares `"php": "8.2 - 8.5"`, so composer refuses to resolve dependencies on 8.6.0-dev ([failing run](https://github.com/goaop/framework/actions/runs/30567085444/job/90954197504)). The experimental lane would then only ever tell us "a dependency hasn't updated its constraint yet" instead of whether the framework itself works on 8.6.

`--ignore-platform-req=php+` (note the `+`) relaxes only the *upper* PHP bound — lower bounds and extension requirements are still enforced. And it's applied only when `matrix.experimental` is true, so the 8.4/8.5 jobs keep strict platform checks. This mirrors what nightly-PHP CI setups commonly do; once dependencies officially allow 8.6 the flag becomes a no-op and can be dropped.

---

PR #588, opened 2026-08-25
Boot-time profiling: span instrumentation + benchmark harness

lisachenko:
## What

First step of the hot-start boot-time optimization work for 4.0: measurement infrastructure.

- **`Go\Instrument\BootTimer`** — a tiny boot profiler that is disabled unless the `GO_AOP_BENCH_SPANS` env var points to an output file (a single static bool check per call when off, so it can live in the hot path permanently). Records hrtime spans with per-span autoloaded-file and memory deltas, plus aggregate counters, and dumps a JSON report on shutdown.
- **Span hooks** around each boot phase: `AspectKernel::init` sub-steps (normalizeOptions, container construction, stream-filter registration, transformer registration, composer-loader init, `configureAop`), per-service `Container::addLazyService`, the `_transformation.cache` include in `CachePathManager`, `AopComposerLoader::findFile` (composer lookup / realpath / cache-hit / filter-rewrite counters), and the transformer cold path (`SourceTransformingLoader::filter` parse, `CachingTransformer` weave + cached-reparse).
- **`tests/Benchmark/`** — `bench-request.php` simulates one request against the fixture project (kernel boot + autoload of woven classes + first intercepted call); `run-bench.php` orchestrates the cold/warm matrix across PHP binaries, using an opcache file cache to emulate a persistent server opcache in CLI runs, and reports median/p10/p90.
- **Fixture project**: new `production` configuration profile (`debug => false`, isolated cache dir) so the warm-cache production fast path can be measured next to the existing debug profile.

## Why

Baseline numbers and a per-phase breakdown are needed before the planned 4.0 boot-time optimizations (lazy transformer registration, truly-lazy container services, lazy aspect registration, real `PREBUILT_CACHE` mode, opcache-friendly cache state). Follow-up commits on this branch will add the measured results and the optimizations themselves.

## Notes

- No behavior change when `GO_AOP_BENCH_SPANS` is not set.
- Benchmark results and analysis will be posted on this PR once the matrix (PHP 8.4/8.5/8.6 × cold/warm × debug/production) has been run.

---

### DIFF
--- a/src/Core/AspectKernel.php
+++ b/src/Core/AspectKernel.php
@@ -14,6 +14,7 @@ namespace Go\Core;
 
 use Go\Aop\AspectException;
 use Go\Aop\Features;
+use Go\Instrument\BootTimer;
 use Go\Instrument\ClassLoading\AopComposerLoader;
 use Go\Instrument\ClassLoading\CachePathManager;
 use Go\Instrument\ClassLoading\SourceTransformingLoader;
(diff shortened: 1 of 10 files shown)


lisachenko:
It isn’t valid direction to commit this into main code and PR. Declining this - only for local development and checks

PR #589, opened 2026-08-25
perf(container): defer service construction with factory map instead of eager lazy proxies

lisachenko:
## What

`Container::addLazyService()` previously executed `new ReflectionClass($id)` + `newLazyProxy()` for each of the 9 built-in services on **every request**, autoloading each service class and its whole parent hierarchy — including the Dissect lexer/parser classes — even on a hot-cache request where none of those services are ever touched.

This change makes registration truly lazy:

- `addLazyService()` now only stores the factory closure — no reflection, no proxy, no autoloading at boot.
- The service is constructed, interface-tagged and resource-tracked on its **first retrieval** (`getService()` / `getValue()`), with a re-entrancy guard.
- `getServicesByInterface()` materializes pending factories that implement the requested interface before answering, so the weaving and console paths observe exactly the same set of services as before. This path never runs on a hot request.
- `has()` accounts for both constructed values and pending factories.

## Why (measured)

Boot profile of the fixture project (`tests/Fixtures/project`, 9 aspects, warm cache, `debug=false`, opcache file cache, median of 15 runs; kernel init + 6 woven class autoloads + first intercepted call):

| | PHP 8.5 master | PHP 8.5 this PR | PHP 8.4 master | PHP 8.4 this PR |
|---|---|---|---|---|
| `AspectKernel::init()` | 1.91 ms | **1.14 ms (−40 %)** | 1.75 ms | **1.08 ms (−38 %)** |
| full request | 3.41 ms | **2.90 ms (−15 %)** | 3.51 ms | **2.90 ms (−17 %)** |
| included files | 142 | **129** | 158 | **145** |

Cold-start behavior is unchanged (weaving dominates there and all services materialize anyway).

## BC note (4.0)

Services registered via `addLazyService()` are no longer PHP 8.4 lazy proxy instances available from boot; they are constructed on first retrieval. Code that relied on `getValue()` returning an uninitialized proxy observes the factory running at retrieval time instead. The container's public API surface is unchanged.

`tests/Core/ContainerTest.php` was updated accordingly: the proxy-specific test became `testLazyServiceIsNotConstructedUntilFirstRetrieval`, which pins the new deferred-construction semantics (registration alone never invokes the factory).

First of five sequential PRs from the hot-start boot profiling effort (container laziness → lazy aspects → lazy transformers → trusted prebuilt cache → cache-state split). Each lands independently; later ones build on merged earlier ones.

---

### DIFF
--- a/src/Core/Container.php
+++ b/src/Core/Container.php
@@ -20,7 +20,6 @@ use Go\Aop\Pointcut\PointcutLexer;
 use Go\Aop\Pointcut\PointcutParser;
 use Go\Instrument\ClassLoading\CachePathManager;
 use OutOfBoundsException;
-use ReflectionClass;
 use ReflectionObject;
 
 /**
@@ -32,6 +31,11 @@ class Container implements AspectContainer
      * @var array<string, mixed> Hashmap of items/services in the container
      */
     private array $values = [];
+
+    /**
+     * @var array<class-string, Closure(AspectContainer): object> Deferred service factories, keyed by class-name
+     */
+    private array $factories = [];
 
     /**
      * @var array<class-string, list<string>> Holds information about mapping of interface tags into identifiers
@@ -122,13 +126,14 @@ class Container implements AspectContainer
 
     final public function addLazyService(string $id, Closure $lazyInitializationClosure): void
     {
-        $reflectionClass = new ReflectionClass($id);
-        $proxy = $reflectionClass->newLazyProxy(fn (object $proxy): object => $lazyInitializationClosure($this));
-        $this->add($id, $proxy);
+        $this->factories[$id] = $lazyInitializationClosure;
     }
 
     final public function getService(string $className): object
     {
+        if (!isset($this->values[$className]) && isset($this->factories[$className])) {
+            $this->materializeService($className);
+        }
         if (!isset($this->values[$className])) {
             throw new OutOfBoundsException("Value $className is not defined in the container");
         }
@@ -142,7 +147,11 @@ class Container implements AspectContainer
     final public function getValue(string $key): mixed
     {
         if (!isset($this->values[$key])) {
-            throw new OutOfBoundsException("Value $key is not defined in the container");
+            if (isset($this->factories[$key])) {
+                $this->materializeService($key);
+            } else {
+                throw new OutOfBoundsException("Value $key is not defined in the container");
+            }
         }
 
         return $this->values[$key];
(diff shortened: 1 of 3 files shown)


lisachenko:
CI status on this head: all required checks are green (PHPStan, PHPUnit lowest/highest on 8.4 and 8.5). The two failing jobs are the **experimental PHP 8.6 jobs** (`continue-on-error: true`), and the failure is not caused by this PR:

`Go\Proxy\Part\InterceptedFunctionGeneratorTest::testGenerate@strcoll` — PHP 8.6.0 added `#[\Deprecated(message: 'use Collator::compare() instead', since: '8.6')]` to `strcoll()`, so the generated function proxy now includes the attribute while the test expects the bare signature. It reproduces identically on any branch under 8.6 (it failed the same way on the unrelated head of #588) and is deterministic, so a re-run would not change it; this PR does not touch function proxy generation.

Proposed fix (separate small PR, since it's out of scope here): in `tests/Proxy/Part/InterceptedFunctionGeneratorTest.php`, make the expectation version-aware, e.g. strip attribute lines before comparing or gate an expanded expected string behind `PHP_VERSION_ID >= 80600`:

```php
$generated = preg_replace('/^#\[\\\\Deprecated\(.*\)\]\n/m', '', $generated);
```

Happy to open that follow-up PR on request.

---

src/Core/Container.php
```diff
@@ -122,13 +126,14 @@ final public function add(string $id, mixed $value): void
 
     final public function addLazyService(string $id, Closure $lazyInitializationClosure): void
     {
-        $reflectionClass = new ReflectionClass($id);
-        $proxy = $reflectionClass->newLazyProxy(fn (object $proxy): object => $lazyInitializationClosure($this));
-        $this->add($id, $proxy);
```
lisachenko, line 127: If we are not calling add(), then we can’t register tags/interface implementation properly. How it influences on the code?

src/Core/Container.php
```diff
@@ -122,13 +126,14 @@ final public function add(string $id, mixed $value): void
...
+        $this->factories[$id] = $lazyInitializationClosure;
```
lisachenko, line 135: Id can be any string here, this may break the logic of tag checking below, when is_subclass_of will be called for random strings

src/Core/Container.php
lisachenko: Tags still end up registered — just later. The lifecycle is: `add()` (with its interface-tag registration and resource tracking) runs inside `materializeService()` on first retrieval instead of at registration. For consumers that query *before* a service was ever retrieved, `getServicesByInterface()` first materializes any pending factory whose class implements the requested interface (the loop above), so every existing consumer — `WeavingTransformer`'s `Advisor` query, `AspectLoader::getUnloadedAspects()`'s `Aspect` query, the debug commands — observes exactly the same result set as with eager registration. `testLazyServiceIsTaggedByInterface` pins this.

The influence on runtime: tag queries only happen on the weaving/console paths, so a hot-cache request never pays for tagging (or for the `ReflectionObject` + `is_readable()` resource tracking) of services it never touches — that's where the measured −0.77 ms/boot comes from.

---

src/Core/Container.php
lisachenko: Good catch — fixed in 0a5df06 (the commit just pushed): `addLazyService()` now rejects ids that aren't syntactically valid class names (checked with a regex, so nothing is autoloaded at registration), enforcing the class-string contract the interface docblock already declares. Arbitrary string keys keep going through `add()` as before, so `$factories` can only hold plausible class names and `is_subclass_of()` never runs on random strings. Added `testLazyServiceRejectsNonClassNameId` to cover it.

(Even before the guard the tag logic itself stayed correct — `is_subclass_of` on a junk id returns false after a failed autoload probe, so the entry was skipped — but the probe itself was wasteful and the failure mode silent; now it's an explicit `AspectException` at registration.)

---

lisachenko: Correction: the commit is 9323b5d (not 0a5df06).

---

lisachenko:
Update on the PHP 8.6 experimental jobs: the fix now exists as #590 (swap the `strcoll` dataset to `strcmp`), and I've ported that one-line test change into this branch as 145edcd so the 8.6 jobs go green here too. It becomes a no-op once #590 merges into master.

---

PR #592, opened 2026-08-25
perf(kernel): register transformation pipeline lazily on first cache miss

lisachenko:
## What

The stream filter and the transformer objects (`CachingTransformer`, `FilterInjectorTransformer`, `MagicConstantTransformer`) were constructed and registered on **every request**, although a warm-cache request never streams a single file through the filter. This PR defers all of it to the first cache miss:

- `AspectKernel::init()` registers `CachingTransformer` as a deferred container definition (riding on the factory map from #589) — no `stream_filter_register()`, no transformer construction at boot.
- New idempotent `SourceTransformingLoader::ensureRegistered($container)` brings up the filter + pipeline from the miss paths.
- `FilterInjectorTransformer::rewrite()` self-configures from the booted kernel on first call — its call sites (autoloader miss path, rewritten `include` statements inside cached files) can run before any transformer object exists. `configure()` became set-once (first configuration wins) instead of throwing on a second call.
- `MagicConstantTransformer::resolveFileName()` — called at runtime from woven code — resolves its paths on demand from the kernel options instead of relying on boot-time constructor side effects.
- `CacheWarmer` calls `ensureRegistered()` before streaming files through the filter URI it builds directly (also covers `debug:weaving`).

Also adds the missing CHANGELOG entries for this and the two previous laziness changes (#589, #591).

## Why (measured)

Fixture project, warm cache, `debug=false`, opcache file cache, paired back-to-back runs (median of 21), PHP 8.5:

| | master (with #589+#591) | this PR |
|---|---|---|
| `AspectKernel::init()` | 1.28 ms | **0.90 ms** |
| full request | 3.27 ms | **2.71 ms (−17 %)** |
| included files | 123 | **117** |

Cumulative since the series started: init 1.88 → 0.90 ms, files 142 → 117 on the same benchmark.

## BC note (4.0)

Protected `AspectKernel::registerTransformers()` is replaced by `createSourceTransformers()`, which returns the inner transformer chain (the `CachingTransformer` wrapper is now owned by the kernel/container). Kernels overriding the old hook must migrate. `FilterInjectorTransformer::configure()` no longer throws when called twice — the first configuration wins.

## Validation

PHPStan clean; 2480 tests green on PHP 8.5, 8.4 and 8.6.0beta2 locally (cold weaving, warmup command and debug commands all exercise the lazy pipeline through the functional tests).

Third of five sequential PRs (container laziness → lazy aspects → **lazy transformers** → trusted prebuilt cache → cache-state split).

---


src/Core/AspectKernel.php
```diff
@@ -275,39 +281,34 @@ protected function normalizeOptions(array $options): array
...
+     * Returns list of source transformers, that will be applied to the PHP source on a cache miss
+     *
+     * Called lazily by the CachingTransformer when the first file actually needs weaving;
+     * never runs on a warm-cache request.
      *
      * @return SourceTransformer[]
      * @internal This method is internal and should not be used outside this project
      */
-    protected function registerTransformers(): array
+    protected function createSourceTransformers(): array
```
lisachenko: Is it possible to move this to container registration as typical services? Tagged and loaded

lisachenko: They can also benefit from lazy init

src/Core/AspectKernel.php
```diff
@@ -275,39 +281,34 @@ protected function normalizeOptions(array $options): array
...
 
     /**
-     * Returns list of source transformers, that will be applied to the PHP source
+     * Returns list of source transformers, that will be applied to the PHP source on a cache miss
+     *
+     * Called lazily by the CachingTransformer when the first file actually needs weaving;
+     * never runs on a warm-cache request.
      *
      * @return SourceTransformer[]
      * @internal This method is internal and should not be used outside this project
```
lisachenko: It can be even fully dropped and moved to container definitions approach

src/Core/AspectKernel.php
lisachenko: Done in d2cffae: every transformer (ConstructorExecution, FilterInjector, Weaving, MagicConstant, Caching) is now a typical deferred container service — nothing constructs at boot, and the chain is assembled by the `SourceTransformer` interface tag in registration order (registration order = transformation order, feature flags gate registration exactly as they gated construction). A custom kernel can plug its own transformer in with a single `addLazyService()` from `configureAop()` and it gets tagged and picked up automatically.

---

src/Core/AspectKernel.php
lisachenko: Done in 4aceb78: `createSourceTransformers()` is gone entirely — the chain assembly (tag query minus the `CachingTransformer` wrapper) lives inside the `CachingTransformer` container definition itself. No protected kernel hook remains; the pipeline is customized purely through container service registration. CHANGELOG updated accordingly (`registerTransformers()` removed rather than renamed).

---

src/Core/AspectKernel.php
lisachenko: Follow-up for downstream libraries: AspectMock's `Kernel` overrode the old `registerTransformers()` to build a fully custom chain (FilterInjector → MagicConstant → its own `BeforeMockTransformer` instead of the standard `WeavingTransformer`), i.e. it needs to replace/omit/reorder built-ins, not just append. 484f6d8 restores that extension point in container form: the built-in registrations moved into a protected `registerTransformerServices(AspectContainer $container)` hook — overriding it controls the pipeline completely (their migration is 1:1: register FilterInjector, MagicConstant and `BeforeMockTransformer::class` definitions; the standard weaver is simply never registered). Re-registering a pending service id also replaces its factory while keeping its chain position (pinned by a new test), and plain appending stays a one-liner from `configureAop()`. CHANGELOG documents the migration.

---

src/Core/AspectKernel.php
```diff
@@ -138,11 +138,29 @@ public function init(array $options = []): void
...
+                $this,
+                static function () use ($container): array {
+                    // Tagged loading: every registered service implementing SourceTransformer
+                    // forms the chain, in registration order. A kernel can plug its own
+                    // transformer in with a single addLazyService() call from configureAop().
+                    $transformers = $container->getServicesByInterface(SourceTransformer::class);
+                    // The caching transformer is the pipeline entry point wrapping this chain
+                    unset($transformers[CachingTransformer::class]);
+
+                    return array_values($transformers);
```
lisachenko: keys can be preserved, no need to use array_values, this will be extra function call without real need

lisachenko: Fixed in 25936b2 — `array_values()` dropped, class-name keys preserved (the chain consumers only iterate values).

---

PR #594, opened 2026-08-26
perf(cache): split runtime include map from transformation metadata

lisachenko:
## What

`AopComposerLoader` materialized the full `_transformation.cache` array (filemtime + cacheUri per file) on **every request**, while the hot path only needs the `originalPath => cachedPath` mapping. The cache state is now two opcache-friendly PHP return-array files:

- **`_include.cache`** — minimal `originalPath => cacheUri|null` map (null = known but untransformed), read by the autoloader on every request via the new `CachePathManager::queryIncludeMap()`; roughly half the data of the old array.
- **`_transformation.cache`** — full build metadata, unchanged format, now loaded **lazily** and only on the cache-miss/weaving paths (`queryCacheState()` / flush). A hot request never materializes it.

Details:
- `flushCacheState()` loads the full metadata before merging (so unqueried entries are never dropped) and writes both files; `setCacheState()` keeps the in-memory include map coherent.
- **Legacy cache dirs keep working**: without `_include.cache`, the map is derived from the metadata once at boot (same cost as before the split) and the next flush/warmup writes both files.
- `findFile()` semantics preserved exactly: production + known file → cached path (or original when untransformed, bypassing the filter); debug/unknown → filter path.

## Measured

Paired back-to-back runs (fixture project, ~30 cached files, warm cache, prod, PHP 8.5, median of 21): init 1.30 → 1.18 ms, full request −0.18 ms (−5 %). The fixture's cache map is small — the cost this removes scales **linearly with project size** (a 10k-file app materializes a 10k-entry × 2-field array per request today; after this PR it's a flat string map, and the metadata array not at all).

## Tests

New `tests/Instrument/ClassLoading/CachePathManagerTest` pins: both files written on flush, include map served without loading the full metadata (verified via the lazy-load flag), full metadata still lazily available, and the legacy-directory fallback. Full suite green on PHP 8.5/8.4 (2486 tests), PHPStan clean.

Fifth and final PR of the series (container laziness → lazy aspects → lazy transformers → trusted prebuilt cache → **cache-state split**).

---


src/Instrument/ClassLoading/AopComposerLoader.php
```diff
@@ -155,12 +155,11 @@ public function findFile(string $class): false|string
             if (is_string($resolved)) {
                 $file = $resolved;
             }
-            $cacheState = $this->cacheState[$file] ?? null;
-            if ($cacheState && $this->isProduction) {
-                $cacheUri = is_array($cacheState) && is_string($cacheState['cacheUri'] ?? null) ? $cacheState['cacheUri'] : null;
-                $file     = $cacheUri ?: $file;
+            if ($this->isProduction && array_key_exists($file, $this->includeMap)) {
```
lisachenko: If it is production and we have already classmap, it can be given to composer’s default autoloader via addClassMap or similar method, please check real API. The only thing required - is to filter null values, to prevent overrides

lisachenko: But this requires also consideration of composer psr roots, to remap plain filenames back to the classes. So, I’m not considering this seriously, just an idea to check, which should be verified.

tests/Instrument/ClassLoading/CachePathManagerTest.php
```diff
@@ -0,0 +1,121 @@
...
+            mkdir(self::$appDir, 0777, true);
+        }
+        if (!is_dir(self::$cacheDir)) {
+            mkdir(self::$cacheDir, 0777, true);
+        }
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        array_map(unlink(...), glob(self::$cacheDir . '/*') ?: []);
```
lisachenko: Very dangerous part of code, one mistake - and important directories will be erased. Together with rmdir calls below.

tests/Instrument/ClassLoading/CachePathManagerTest.php
```diff
@@ -0,0 +1,121 @@
...
+        @rmdir(self::$cacheDir);
+        @rmdir(self::$appDir);
+    }
+
+    protected function setUp(): void
+    {
+        array_map(unlink(...), glob(self::$cacheDir . '/*') ?: []);
```
lisachenko: Same here, very dangerous part of code, one mistake and some files will be erased

src/Instrument/ClassLoading/CachePathManager.php
```diff
@@ -27,10 +27,15 @@
...
     /**
-     * Name of the file with cache paths
+     * Name of the file with full transformation metadata (build-time data, loaded lazily)
      */
     private const CACHE_FILE_NAME = '/_transformation.cache';
 
+    /**
+     * Name of the file with the minimal runtime include map (originalPath => cacheUri|null)
+     */
+    private const INCLUDE_MAP_FILE_NAME = '/_include.cache';
```
lisachenko, line 37: Is it possible somehow to keep only one such map to avoid sync between two files? Ideally this file should be integrateable right into the composer loader. Or via simple transformation. Write comment and receive confirmation before fixing

tests/Instrument/ClassLoading/CachePathManagerTest.php
lisachenko: Hardened in 78ab46f: the cleanup no longer glob-sweeps — it asserts the directory value starts with `sys_get_temp_dir() . '/goaop-cpm-'` and then removes only the two exact file names this test writes (`_transformation.cache`, `_include.cache`). A wrong directory value now fails the assertion instead of deleting anything. Applied the same guard to `CachedAspectLoaderTest` (from #593), which used the same pattern.

---

tests/Instrument/ClassLoading/CachePathManagerTest.php
lisachenko: Same fix — 78ab46f: prefix-asserted, known-file-only removal; the per-test `setUp()` uses the identical guarded helper.

---

src/Instrument/ClassLoading/CachePathManager.php
lisachenko: Proposal (not implementing until you confirm):

**Option A — class-keyed runtime map, integrated into the composer loader (recommended).** `WeavingTransformer` knows the FQCN of every class it weaves, so record it in the cache metadata at build time. The single runtime file then becomes class-keyed:

```php
// _include.cache
<?php return [
    'map'  => ['App\Foo' => AOP_CACHE_DIR . '/src/Foo.php', ...],  // transformed classes only
    'skip' => ['App\Bar' => true, ...],                            // known untransformed classes
];
```

At production boot: `$originalComposerLoader->addClassMap($map)` — verified against the real API: `Composer\Autoload\ClassLoader::addClassMap(array)` exists and merges, and `findFile()` consults `$this->classMap` **before** PSR-4, so woven classes resolve natively through composer with zero wrapper work. Untransformed classes aren't added to the classmap (your "filter null values" point — composer resolves them to the original file anyway); the `skip` set lets the wrapper return composer's result untouched instead of re-running the include-path filter. Unknown classes still fall through to the wrapper's transform-on-miss path.

Because the class names are recorded at weaving time, **no psr-root reverse-mapping is needed** (the concern from your other comment). Bonus: the wrapper's per-class `PathResolver::realpath()` becomes unnecessary for known classes — the class name is the key, not the resolved path.

Sync-wise this stays one runtime file projected from `_transformation.cache` in the same `flushCacheState()` write; the metadata file remains the single build-time source of truth. Fully collapsing to literally one file would mean materializing the full metadata on every request again, which is what this PR removes.

**Option B — keep this PR's file-keyed map as-is** and treat A as a follow-up PR (it touches WeavingTransformer + cache format again).

Which way do you want it — A inside this PR, A as a follow-up, or stay with B?

---

src/Instrument/ClassLoading/AopComposerLoader.php
lisachenko: Verified — the idea is sound and the psr-root concern is avoidable: `ClassLoader::addClassMap(array $classMap)` exists (merges into `$classMap`) and `findFile()` checks the classmap **before** PSR-4/PSR-0 resolution, so class => cacheUri entries short-circuit natively. The key requirement is class-keyed data, and we can get it without reversing file paths through psr roots: `WeavingTransformer` already holds the FQCN of every class it weaves, so the class name can be recorded into the cache state at build time. Untransformed entries simply stay out of the classmap (no overrides), matching your "filter null values" note. Full sketch + open question posted on the `CachePathManager` thread — awaiting your confirmation before implementing.

---

src/Instrument/ClassLoading/CachePathManager.php
lisachenko: A inside this PR

lisachenko: Implemented as Option A in 87f8ef6: `WeavingTransformer` records every discovered FQCN into the cache metadata; `_include.cache` is now `['map' => [class => cachedFile], 'skip' => [class => true]]`, and at production boot the map is handed to composer via `addClassMap()` — woven classes resolve natively (composer checks the class map before PSR-4), untransformed classes stay out of the map and are served untouched, and the wrapper's per-class `realpath()` disappears for known classes. Pre-class-map cache dirs are treated as stale (one re-weave, or `cache:warmup:aop` at deploy) — documented in CHANGELOG. Verified end-to-end on the fixture app: generated map is composer-ready, class autoload time in the hot request dropped measurably (paired runs), full suite + PHPStan green on 8.5/8.4.

---

PR #595, opened 2026-08-27
Move caching into SourceTransformingLoader, remove CachingTransformer

lisachenko:
## Summary

Follow-up to the boot-time series (#589, #591, #592, #593, #594), as requested: `CachingTransformer` is removed and the caching logic lives right in `SourceTransformingLoader::filter()`, next to the `StreamMetaData` construction. If a usable cache record exists for the streamed file, its content is returned from there as the result — otherwise the code falls back to `transformCode()`.

## What changed

- **`SourceTransformingLoader::filter()`** decides in this order:
  1. *Guard* — no cache path, or cache path equals the original file: the source passes through untouched (previously the `RESULT_ABORTED` branch of the wrapper).
  2. *Cache hit* (`findCachedContent()`) — the cached file content (or, for a known-untransformed file, the buffered original source) is emitted **as-is: no `StreamMetaData`, no parsing, no transformer is ever constructed**. All freshness rules are ported verbatim: `filemtime` comparison, recorded-`cacheUri` match, `hasAnyResourceChangedSince()`, and the `PREBUILT_CACHE` trusted-record short-circuit.
  3. *Cache miss* — parse into `StreamMetaData`, run the transformer chain, persist via `saveToCache()` (ported verbatim: `__AopProxied` suffix rename, `mkdir`/`LOCK_EX`/`chmod`, `setCacheState()` with the class-map recording from #594 intact).
- **`transformCode()`** now returns the overall `TransformerResultEnum` (the logic formerly in `CachingTransformer::processTransformers()`); the chain is assembled lazily on the first miss from the tagged container services (`getServicesByInterface(SourceTransformer::class)`), so no `unset()` exclusion is needed anymore.
- **Removed** `src/Instrument/Transformer/CachingTransformer.php`, its container definition in `AspectKernel::init()`, and `SourceTransformingLoader::addTransformer()`. The `registerTransformerServices()` extension hook is unchanged.

## Why it helps

The old cached branch went through the whole filter pipeline anyway and then **re-parsed the cached file** (`ReflectionEngine::parseFile()`) just to rebuild the token stream. Now a hit skips parsing entirely — which finally makes the warm-cache fast path apply in **debug mode** too, with unchanged invalidation semantics.

Local numbers (fixture app, opcache file cache, median of 21):

| Mode (PHP 8.5) | master | this PR | Δ |
|---|---|---|---|
| debug warm boot+request | 15.1–16.2 ms | 5.7–6.0 ms | **−62 %** |
| debug warm files included | 187 | 127 | −60 files |
| production warm | 3.9 ms | 3.9 ms | unchanged |
| cold (first boot) | ~103 ms | ~103 ms | unchanged |

PHP 8.4 debug warm shows the same shape: 16.2 ms → 5.5 ms (−66 %).

## Validation

- `phpunit` green on PHP 8.5 and 8.4 (2486 tests), including the functional warmup / debug:weaving / inconsistent-weaving coverage that exercises the moved logic on real caches
- `phpstan` (level max) clean

---


lisachenko:
i do not see any tests here - please cover with tests

lisachenko:
Added in fd84b7e: `tests/Instrument/ClassLoading/SourceTransformingLoaderTest.php` covers every branch of the filter's cache decision through the real stream filter (`php://filter/read=.../resource=...`), in separate processes because of the loader's static state:

- fresh **transformed** record → cached file content served, `getServicesByInterface()` never called (no transformer constructed)
- fresh **untransformed** record → original source served, same no-transformer guarantee
- **miss** → chain runs, woven file written, cache state recorded with its `cacheUri`
- **stale** record (older than the source) → falls back to the chain
- **abstaining** chain → recorded as untransformed, no cache file written
- **aborting** transformer → rest of the chain skipped, nothing written
- **`PREBUILT_CACHE`** → deliberately stale record trusted, `hasAnyResourceChangedSince()` never consulted
- **guard** (cache path == original file) → source passes through untouched, no transformers, no state written

Full suite green on PHP 8.5/8.4 (2494 tests), PHPStan clean.

---

PR #597, opened 2026-08-28
PHP 8.5 feature-support audit: harness and fixtures

lisachenko:
## PHP 8.5 feature-support audit

Audit branch: harness + fixtures that weave PHP 8.0–8.5 feature constructs through the real `WeavingTransformer` and lint the woven trait and generated proxies. Run on **PHP 8.4, 8.5.10, and 8.6.0beta2** — results identical across versions (nothing is 8.5-runtime-specific; no 8.6-specific breakage found anywhere in the suite).

### Verified working on PHP 8.5
- Pipe operator `|>` in woven method bodies
- `clone with`
- `#[\NoDiscard]` (copied to proxy; join-point dispatch preserves the value path)
- Attributes on class constants, incl. `#[\Deprecated]`
- Closures / first-class callables as parameter defaults (support merged earlier)
- `final` promoted properties and static asymmetric visibility on non-intercepted properties
- `self`/`parent` reflection resolution

### Confirmed gaps → filed issues
| Issue | Bug | Feature since |
|---|---|---|
| #598 | Class-level attributes break class→trait conversion (parse error) | 8.0 |
| #599 | Promoted ctor properties × property interception → broken constructor | 8.0/8.4 |
| #600 | Enum proxies lose constant-expression case values (fatal) | 8.1 |
| #601 | Non-scalar / closure attribute arguments crash proxy generation | 8.1/8.5 |
| #602 | Global constants in attribute args crash weaving (upstream parser-reflection) | 8.0 |
| #603 | `ConstructorExecutionTransformer` rewrites `new` in constant expressions (fatal) | 8.1 |

Enhancements/docs/modernization: #604 (pointcut expressiveness), #605 (php85-limitations docs), #606 (native lazy objects), #607 (property hooks internally), #608 (typed constants + `#[\Override]`), #609 (array_any/find + latent strpos bugs), #610 (process/cleanup).

### How the harness works
`tests/Instrument/Transformer/Php85AuditScratchTest.php` weaves every fixture in `tests/Fixtures/audit/src/` with an intercept-everything advice matcher (mirroring real `AdviceMatcher` gates) and lints all outputs. Confirmed-broken fixtures are pinned in `KNOWN_GAPS` with their issue link: CI stays green, and once a fix PR lands, the corresponding entry must be removed (the harness fails loudly when a known gap starts passing).

### Fix work
Fixes are being prepared as separate PRs per workstream (attribute AST fidelity; weaving-transformer token fixes; enum/pointcuts/docs; modernization batch).


composer.json
```diff
@@ -52,7 +52,8 @@
         },
         "psr-4": {
             "Go\\": "tests/",
-            "Go\\Tests\\TestProject\\": "tests/Fixtures/project/src/"
+            "Go\\Tests\\TestProject\\": "tests/Fixtures/project/src/",
+            "Go\\Tests\\Audit\\": "tests/Fixtures/audit/src/"
```
lisachenko: Would be nice to reuse existing folders instead of creating the special audit one. Put stubs/tests somewhere else

lisachenko: Done in 69c36d3 — the audit stubs now live in `tests/Stubs/Audit/` (covered by the existing `Go\` → `tests/` mapping), and the extra `composer.json` autoload-dev entry is gone. Full suite + PHPStan re-verified green on 8.4/8.5.

---

tests/Stubs/Collaborator.php
lisachenko, line 1: I don't like the Audit namespace - simply put into the Stubs directory all files

lisachenko: Done — all 16 stub files now sit directly in `tests/Stubs/` under `Go\Stubs`, the `Audit/` directory and sub-namespace are gone, and the harness picks its fixtures from an explicit list instead of globbing. Full suite + PHPStan green on 8.4/8.5.

---

PR #613, opened 2026-08-28
Fix weaving of class attributes, promoted properties, and new-in-initializers

### LINKED ISSUE #598: Class-level attributes break class→trait conversion (woven trait has parse error)
## Summary

Weaving any class that carries a **class-level attribute** produces a corrupted trait: `WeavingTransformer::convertClassToTrait()` mangles the attribute tokens and drops the `class` keyword entirely, so the woven file no longer parses.

Found during the PHP 8.5 support audit (PR #597). Reproduces identically on PHP 8.4, 8.5.10 and 8.6.0beta2.

## Reproduction

```php
#[ExprAttr]
class Php80ClassAttrPlain
{
    public function run(): int
    {
        return 42;
    }
}
```

Woven output (method advice on `run`):

```php
#[Php80ClassAttrPlain__AopProxied
 
{
    public function run(): int
    ...
```

```
PHP Parse error: syntax error, unexpected token "{", expecting "]"
```

With attribute arguments it gets worse — `#[\Attribute(\Attribute::TARGET_ALL)]` becomes `#[\Attribute(\Attribute::Php80ClassAttrPlain__AopProxied` (the first `T_STRING` inside the attribute group is renamed to the trait name).

Fixture: `tests/Fixtures/audit/src/Php80ClassAttrPlain.php`, harness `tests/Instrument/Transformer/Php85AuditScratchTest.php` on the audit branch.

## Root cause

`src/Instrument/Transformer/WeavingTransformer.php` — `convertClassToTrait()` (~line 237) starts its token scan from the class node's start position. In nikic/php-parser, a `ClassLike` node's `startTokenPos` **includes its attribute groups**, so the scan that expects `final|abstract|readonly` modifiers followed by `T_CLASS` and the class name instead lands inside the `#[...]` group: the first `T_STRING` there gets renamed to `<Class>__AopProxied`, and the delete-until-`{` step then eats the rest of the attribute plus the real `class <Name>` header.

Note the enum path (`convertEnumToTrait()`) is not affected — the existing `#[Loggable] enum BackedEnum` functional fixture weaves fine — which is why this was never caught: no existing woven-class fixture has a class-level attribute.

## Proposed fix

In `convertClassToTrait()`, begin the token scan after the class node's attribute groups — e.g. use the position after the last `AttributeGroup` sub-node (or scan forward for the actual `T_CLASS` token at depth 0, skipping `T_ATTRIBUTE`…`]` ranges the way `stripOverrideAttributeFromInterceptedMethods()` already tracks attribute-group extents). Class-level attributes should remain on the woven trait untouched (attributes are legal on traits) — or be dropped there and preserved on the proxy, mirroring the `readonly` handling; preserving them on the trait is the smaller change.

Add golden fixtures: attribute-bearing class with and without arguments.

## Environment

- goaop/framework master (4.0-dev), nikic/php-parser 5.8.0, goaop/parser-reflection 4.0.0
- PHP 8.4 / 8.5.10 / 8.6.0beta2 — identical failure

### LINKED ISSUE #599: Property interception breaks promoted constructor properties (parse error or silently lost assignments)
## Summary

When a property pointcut matches a **promoted constructor property**, `WeavingTransformer::commentOutInterceptedPropertiesInTraitBody()` comments the promoted parameter out **inside the constructor signature**. Depending on formatting this yields either a parse error or a constructor that silently discards its arguments.

Found during the PHP 8.5 support audit (PR #597). Reproduces on PHP 8.4 / 8.5.10 / 8.6.0beta2.

## Reproduction

Single-line constructor → parse error (the `// ` prefix swallows the closing `)`):

```php
class Collaborator
{
    public function __construct(public string $tag = 'default') {}
}
```

woven trait:

```php
public function __construct(// public string $tag = 'default' // Moved by weaving interceptor to the {@see ...->tag})
{
}
```

Multi-line constructor → lints, but is semantically broken:

```php
trait Php85CloneWith__AopProxied
{
    public function __construct(
        // public string $name = 'initial' // Moved by weaving interceptor ...,
        // public int $count = 0 // Moved by weaving interceptor ...,
    ) {
    }
```

The proxy re-declares the properties with interception hooks and its constructor keeps the parameter list (`__construct(string $name = 'initial', int $count = 0)`), but it delegates to the trait's `__aop____construct(...)` — whose parameters no longer exist. Arguments are silently dropped, the promoted assignment never happens, and any read (`$this->name`) hits an uninitialized typed property. The hook properties on the proxy also lose their default values (`public string $name {` instead of `public string $name = 'initial' {`).

Same failure mode with PHP 8.5 `final` promoted properties (`final public string $identity = 'anon'`), fixture `tests/Fixtures/audit/src/Php85FinalPromotionAsymStatic.php`.

## Root cause

- `src/Instrument/Transformer/WeavingTransformer.php` — `commentOutInterceptedPropertiesInTraitBody()` (~line 550) treats a promoted param like a property declaration statement and comments out its token range inside the parameter list.
- `src/Proxy/Part/InterceptedPropertyGenerator.php` builds the hook property without carrying over the promoted parameter's default.

## Proposed fix

For an intercepted promoted property, **demote instead of remove** in the woven trait: rewrite `public string $name = 'initial'` to a plain parameter `string $name = 'initial'` and inject `$this->name = $name;` at the top of the constructor body (line numbers can be preserved by doing it within the same line/token budget, or accepting the proxy include boundary). The proxy-side hook property should carry the original default value. Alternatively (smaller but restrictive): exclude promoted properties from `access()` interception in `AdviceMatcher` the way static/readonly/hooked properties already are, and document the limitation — but that silently narrows pointcut semantics, so demotion is preferred.

## Environment

- goaop/framework master (4.0-dev), PHP 8.4 / 8.5.10 / 8.6.0beta2

### LINKED ISSUE #603: ConstructorExecutionTransformer rewrites `new` inside constant expressions (fatal on load)
## Summary

With `Features::INTERCEPT_INITIALIZATIONS` enabled, `ConstructorExecutionTransformer` rewrites **every** `New_` node in a file — including `new` in initializers (PHP 8.1): parameter defaults, static variable initializers, and attribute arguments. A method call is not a valid constant expression, so the transformed file is a compile-time fatal.

Found during the PHP 8.5 support audit (PR #597). Reproduces on PHP 8.4 / 8.5.10 / 8.6.0beta2.

## Reproduction

Input (`tests/Fixtures/audit/src/Php81NewInInitializers.php`):

```php
public function run(Collaborator $helper = new Collaborator('method-default')): string
{
    static $memo = new \ArrayObject();
    ...
```

Transformed output:

```php
public function run(Collaborator $helper = \Go\Instrument\Transformer\ConstructorExecutionTransformer::getInstance()->{Collaborator::class}('method-default')): string
{
    static $memo = \Go\Instrument\Transformer\ConstructorExecutionTransformer::getInstance()->{\ArrayObject::class}();
```

```
PHP Fatal error: Constant expression contains invalid operations
```

## Root cause

`src/Instrument/Transformer/ConstructorExecutionTransformer.php:60-88` — the `FindingVisitor` selects all `Expr\New_` nodes with no context filter.

## Proposed fix

Skip `New_` nodes that appear in constant-expression context: while traversing, exclude nodes whose ancestors include `Param->default`, `StaticVar->default`, `Const_`/`ClassConst`, `PropertyItem->default`, `EnumCase->expr`, or `Attribute` args (a small `NodeVisitor` tracking these containers, or checking node attributes of the enclosing structure during the finding pass). Such `new` expressions run inside engine-managed initializer evaluation and cannot be intercepted — document that limitation (they still hit interception when the class itself is woven elsewhere at normal call sites).

## Environment

- goaop/framework master (4.0-dev), PHP 8.4 / 8.5.10 / 8.6.0beta2

lisachenko:
## Summary

Fixes three confirmed weaving bugs found by the PHP 8.5 audit harness:

Fixes #598 / Fixes #599 / Fixes #603

### #598 — Class-level attributes broke class→trait conversion

**Root cause:** a `ClassLike` node's `startTokenPos` in php-parser includes its attribute groups, so `convertClassToTrait()` / `convertEnumToTrait()` / `adjustOriginalTrait()` started scanning inside `#[...]`: the first `T_STRING` of the attribute was renamed to the `__AopProxied` trait name, and the delete-until-`{` step then removed the attribute remainder together with the real class header, producing invalid output like `#[Foo__AopProxied {`.

**Fix:** the scan now starts right after the node's last attribute group (`WeavingTransformer::getPositionAfterAttributeGroups()`). Class-level attributes stay untouched on the woven trait, where they are legal. This also repairs the enum path — the previously "passing" `#[Loggable] enum BackedEnum` functional fixture actually produced a woven trait that failed `php -l`; it now lints cleanly.

### #599 — Property interception broke promoted constructor properties

**Root cause:** `commentOutInterceptedPropertiesInTraitBody()` line-commented the matched promoted parameter inside the constructor signature. Single-line constructors became parse errors (the comment swallowed the closing `)`), and multi-line constructors silently lost their parameters: the proxy kept the original signature and delegated to a now-parameterless trait constructor, dropping all arguments and leaving typed properties uninitialized. The proxy hook property also lost its default (`public string $name {` instead of `public string $name = 'initial' {`), because promoted properties report `hasDefaultValue() === false`.

**Fix (demotion, as suggested in the issue):**
- The woven trait rewrites the promoted parameter to a plain parameter — only the promotion modifiers (`public/protected/private`, `public(set)/protected(set)/private(set)`, `readonly`, `final`) are removed; attributes, type, and default are kept.
- `$this->prop = $prop;` assignments are injected on the opening-brace line of the constructor body (line numbers preserved), routing construction-time writes through the proxy's generated set hook.
- `AbstractInterceptedPropertyGenerator` now reads the default from the `Param` AST node, so the proxy hook property carries the original default value.

All modifier combinations demote cleanly, including PHP 8.4 asymmetric visibility (`public private(set) int $v = 1`) and PHP 8.5 `final` promoted parameters (`final public string $x = 'y'`).

### #603 — `ConstructorExecutionTransformer` rewrote `new` in constant expressions

**Root cause:** the finding pass collected every `New_` node, including PHP 8.1 "new in initializers" occurrences (parameter defaults, static variable initializers, attribute arguments, global constants). The rewritten `...getInstance()->{Foo::class}(...)` form is not a valid constant expression → compile-time fatal error.

**Fix:** a dedicated `NewExpressionFinderVisitor` tracks entry/exit of constant-expression subtrees (`Param`/`StaticVar`/`PropertyItem`/`Const_` initializers, `EnumCase` values, whole `Attribute` nodes) and only collects `new` expressions outside of them. Runtime code nested inside those containers (e.g. property hook bodies on promoted parameters) is still rewritten.

## Tests

- New golden fixtures: `php80-class-attribute` (plain + argument attributes), `php80-promoted-property` (multi-line ctor with defaults, asymmetric visibility, and a non-intercepted promoted param that must stay promoted), `php80-promoted-property-single-line` — each asserting woven trait and generated proxy.
- PHP 8.5-gated (`#[RequiresPhp('>= 8.5.0')]`) test for `final` promoted parameter demotion via a stub that is never eagerly loaded.
- Functional coverage: new `PromotedPropertyInterceptAspect` + `ClassWeavingTest::testPromotedPropertyWeaving` weave the promoted-property classes with the real matcher on every functional run.
- `ConstructorExecutionTransformerTest` provider extended with 8 const-expr cases (defaults, static vars, global const, attribute args, nested `new`, mixed same-file rewrites, hook bodies).

## Test evidence

- `php8.5 vendor/bin/phpunit` — OK, 2507 tests / 2991 assertions (1 pre-existing deprecation)
- `php8.4 vendor/bin/phpunit` — OK, 2507 tests / 2985 assertions (2 skipped: PHP 8.5-gated)
- `php8.6 vendor/bin/phpunit` (8.6.0beta2, informational) — OK, 2507 tests / 2991 assertions
- `php8.5 vendor/bin/phpstan analyze` — level 10, no errors, no new baseline entries
- Runtime smoke check through a real `AspectKernel`: `new PromotedPropertyClass('  x  ', 5)` → `getName() === 'x'`, `counter === 5`; defaults `'initial'`/`1` applied when omitted; `SingleLinePromotedClass` default/custom tag both correct; instances implement `Go\Aop\Proxy`.

---

### DIFF
--- a/src/Instrument/Transformer/ConstructorExecutionTransformer.php
+++ b/src/Instrument/Transformer/ConstructorExecutionTransformer.php
@@ -14,11 +14,8 @@ namespace Go\Instrument\Transformer;
 
 use Go\Aop\Framework\ReflectionConstructorInvocation;
 use Go\Aop\InitializationAware;
-use PhpParser\Node;
-use PhpParser\Node\Expr\New_;
 use PhpParser\Node\Name;
 use PhpParser\NodeTraverser;
-use PhpParser\NodeVisitor\FindingVisitor;
 
 /**
  * Transforms the source code to add an ability to intercept new instances creation
@@ -57,15 +54,16 @@ final class ConstructorExecutionTransformer implements SourceTransformer
      */
     public function transform(StreamMetaData $metadata): TransformerResultEnum
     {
-        $newExpressionFinder = new FindingVisitor(fn(Node $node) => $node instanceof New_);
+        // Skips `new` inside constant-expression contexts (parameter defaults, static var
+        // initializers, attribute arguments, constants, enum cases) — see issue #603.
+        $newExpressionFinder = new NewExpressionFinderVisitor();
 
         // TODO: move this logic into walkSyntaxTree(Visitor $nodeVistor) method
         $traverser = new NodeTraverser();
         $traverser->addVisitor($newExpressionFinder);
         $traverser->traverse($metadata->syntaxTree);
 
-        /** @var Node\Expr\New_[] $newExpressions */
-        $newExpressions = $newExpressionFinder->getFoundNodes();
+        $newExpressions = $newExpressionFinder->getFoundNewExpressions();
 
         if (empty($newExpressions)) {
             return TransformerResultEnum::RESULT_ABSTAIN;
--- a/src/Instrument/Transformer/NewExpressionFinderVisitor.php
+++ b/src/Instrument/Transformer/NewExpressionFinderVisitor.php
@@ -1,0 +1,104 @@
+<?php
+
+declare(strict_types=1);
+/*
+ * Go! AOP framework
+ *
+ * @copyright Copyright 2026, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Go\Instrument\Transformer;
+
+use PhpParser\Node;
+use PhpParser\Node\Attribute;
+use PhpParser\Node\Const_;
+use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Param;
+use PhpParser\Node\PropertyItem;
+use PhpParser\Node\StaticVar;
+use PhpParser\Node\Stmt\EnumCase;
+use PhpParser\NodeVisitorAbstract;
+
+/**
+ * Finds all `new` expressions that are legal to rewrite into runtime interceptor calls.
+ *
+ * Since PHP 8.1 `new` may appear inside constant-expression contexts: parameter default
+ * values, static variable initializers, attribute arguments and global constants (and
+ * php-parser also accepts it in property/class-constant defaults and enum case values).
+ * Such occurrences must stay untouched — the interceptor rewrite
+ * `...getInstance()->{Foo::class}(...)` is not a valid constant expression and would
+ * trigger a compile-time fatal error (https://github.com/goaop/framework/issues/603).
+ */
+final class NewExpressionFinderVisitor extends NodeVisitorAbstract
+{
+    /**
+     * @var list<New_>
+     */
+    private array $newExpressions = [];
+
+    /**
+     * Object ids of subtree roots that are constant-expression contexts
+     *
+     * @var array<int, true>
+     */
+    private array $constExprRoots = [];
+
+    /**
+     * How deep the traversal currently is inside constant-expression subtrees
+     */
+    private int $constExprDepth = 0;
+
+    /**
+     * @return list<New_> All found `new` expressions outside of constant-expression contexts
+     */
+    public function getFoundNewExpressions(): array
+    {
+        return $this->newExpressions;
+    }
+
+    public function enterNode(Node $node): null
+    {
+        if ($node instanceof Attribute) {
+            // Attribute arguments are always constant expressions — skip the whole attribute.
+            // Property hooks on promoted parameters contain runtime code, so for the other
+            // containers only the initializer child expression is marked, not the container.
+            $this->constExprRoots[spl_object_id($node)] = true;
+        } else {
+            $constExpr = match (true) {
+                $node instanceof Param        => $node->default,
+                $node instanceof StaticVar    => $node->default,
+                $node instanceof PropertyItem => $node->default,
+                $node instanceof Const_      => $node->value,
+                $node instanceof EnumCase     => $node->expr,
+                default                       => null,
+            };
+            if ($constExpr !== null) {
+                $this->constExprRoots[spl_object_id($constExpr)] = true;
+            }
+        }
+
+        if (isset($this->constExprRoots[spl_object_id($node)])) {
+            ++$this->constExprDepth;
+        }
+
+        if ($this->constExprDepth === 0 && $node instanceof New_) {
+            $this->newExpressions[] = $node;
+        }
+
+        return null;
+    }
+
+    public function leaveNode(Node $node): null
+    {
+        $nodeId = spl_object_id($node);
+        if (isset($this->constExprRoots[$nodeId])) {
+            --$this->constExprDepth;
+            unset($this->constExprRoots[$nodeId]);
+        }
+
+        return null;
+    }
+}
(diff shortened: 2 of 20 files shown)


lisachenko:
Rebase please

lisachenko:
Rebased onto current master (`c52e072`). One conflict in `WeavingTransformerTest.php` (both #611 and this branch added a test at the same location) — resolved by keeping both tests. Full suite green on 8.4 (`OK, 2523 tests, 4 skipped`) and 8.5 (`OK, 2523 tests, 3040 assertions`, zero deprecations), PHPStan level 10 clean.

---

Issue #610, opened 2026-08-28
Process/cleanup: dead PARAMETER_WIDENING flag, orphaned test fixtures, lost golden coverage, PHPStan on 8.5 in CI


lisachenko:
## Summary

Process and hygiene items from the PHP 8.5 support audit (PR #597); user-approved batch.

## 1. Dead feature flag `PARAMETER_WIDENING`

`src/Aop/Features.php:49-54` gates "parameter widening for PHP>=7.2.0" — unconditional engine behavior for eight years and dead configuration on a ^8.4 floor. The `$useParameterWidening`/`$useTypeWidening` plumbing threads through `WeavingTransformer`, all four proxy generators, `InterceptedConstructorGenerator`, and `MethodGenerator::fromReflection`. Remove flag + plumbing (BC note for 4.0 changelog; `Features` bitmask already has holes at 8/16/32).

## 2. Orphaned golden fixtures — and lost version coverage

14 of 40 files in `tests/Instrument/Transformer/_files/` are referenced by no test (leftovers of the removed `SelfValueTransformer` and `CachingTransformer`):

```
anonymous-class{,-transformed}.php
file-with-self{,-transformed}.php, file-with-self-no-namespace{,-transformed}.php
php80-file{,-transformed}.php
php81-file{,-transformed}.php
php82-file{,-transformed}.php   (input and expected output byte-identical)
yii_style{,_output}.php
```

Deleting them is not enough: their removal from the test flow silently dropped all golden-file coverage of *general* PHP 8.0–8.2 syntax through the weaver (only enums, readonly classes, `#[\Override]`, and property hooks survive). Replace with current-pipeline fixtures covering 8.0–8.2 syntax, 8.3 typed constants, and the audited 8.5 constructs (the audit fixtures in `tests/Fixtures/audit/src/` are the seed).

## 3. PHPStan CI runs only on PHP 8.4

`.github/workflows/phpstan.yml` pins 8.4 while phpunit runs 8.4/8.5(/8.6 experimental). Reflection signatures/semantics differ per version (the 8.5 `self`/`parent` resolution change proved it). Add an 8.5 job.

## 4. Minor

- `tests/functions.php` glob-hijack helper: add parameter/return types.
- Three hand-rolled "static singleton configured on demand" patterns (`FilterInjectorTransformer` `extract()` at :105-109, `MagicConstantTransformer` statics, `SourceTransformingLoader` statics) — at minimum replace the `extract()` with typed reads.
[...]

Issue #602, opened 2026-08-28
Global constants in attribute arguments crash weaving (parser-reflection namespace resolution)


lisachenko:
## Summary

Weaving a namespaced class whose attribute references a **global constant** (`PHP_INT_MAX`, `PHP_EOL`, …) aborts the whole transform:

```
Go\ParserReflection\ReflectionException: Namespace  was not found in the file .../Php80GlobalConstAttrArg.php
```

The root cause is upstream in **goaop/parser-reflection 4.0.0**, but it takes down the framework's weaving pass, so tracking it here from the PHP 8.5 support audit (PR #597). Reproduces on PHP 8.4 / 8.5.10 / 8.6.0beta2.

## Reproduction

```php
namespace Some\Ns;

class Php80GlobalConstAttrArg
{
    #[ExprAttr(PHP_INT_MAX)]
    public function limited(): int { return PHP_INT_MAX; }
}
```

```
#0 vendor/goaop/parser-reflection/src/ReflectionEngine.php:307 parseFileNamespace()
#1 vendor/goaop/parser-reflection/src/ReflectionFileNamespace.php(102)
#2 vendor/goaop/parser-reflection/src/Resolver/NodeExpressionResolver.php(588): resolveExprConstFetch()
#3 vendor/goaop/parser-reflection/src/Resolver/NodeExpressionResolver.php(189)
#4 vendor/goaop/parser-reflection/src/Traits/AttributeResolverTrait.php(63)
#5 src/Proxy/Generator/MethodGenerator.php(135): ReflectionMethod->getAttributes()
```

`NodeExpressionResolver::resolveExprConstFetch()` falls back to looking the constant up in the file's **global namespace** (`ReflectionFileNamespace($file, '')`), and since the file only declares `Some\Ns`, the lookup itself throws.

Fixture: `tests/Fixtures/audit/src/Php80GlobalConstAttrArg.php` (audit branch). This is also why the audit fixtures initially broke `DebugAdvisorCommandTest` — one poisoned file aborts `debug:advisor` for the whole project.

## Proposed fix

Two layers:
1. **Framework**: this disappears for the weave-time path once attribute arguments are copied from the AST instead of evaluated — see #601. That also stops one bad file from aborting the entire transform for this class of error.
[...]

Issue #601, opened 2026-08-28
Non-scalar attribute arguments (enum cases, objects, PHP 8.5 closures) crash proxy generation


lisachenko:
## Summary

`AttributeGroupsGenerator` copies attributes to the proxy by feeding **runtime** `ReflectionAttribute::getArguments()` values into php-parser's `BuilderFactory::args()` → `BuilderHelpers::normalizeValue()`, which only accepts scalars/arrays. Any attribute argument that evaluates to an object kills the whole weave with `LogicException: Invalid value`.

Affected argument kinds:
- enum cases (PHP 8.1): `#[RichAttr(Status::Active)]`
- `new` in initializers (PHP 8.1): `#[RichAttr(new \ArrayObject([1, 2]))]`
- closures / first-class callables in constant expressions (PHP 8.5): `#[ExprAttr(strlen(...))]`, `#[ExprAttr(static function (int $x): int { return $x * 2; })]`

Found during the PHP 8.5 support audit (PR #597). Reproduces on PHP 8.4 / 8.5.10 / 8.6.0beta2 (closure cases are 8.5+ syntax).

## Reproduction

```
TRANSFORM ERROR for Php81NonScalarAttributeArgs: Invalid value

LogicException: Invalid value in vendor/nikic/php-parser/lib/PhpParser/BuilderHelpers.php:276
#0 PhpParser\BuilderFactory->args()
#1 src/Proxy/Generator/AttributeGroupsGenerator.php(59)
#2 src/Proxy/ClassProxyGenerator.php(153)
#3 src/Instrument/Transformer/WeavingTransformer.php(167)
```

Fixtures: `tests/Fixtures/audit/src/Php81NonScalarAttributeArgs.php`, `tests/Fixtures/audit/src/Php85ClosuresInConstExpr.php` (audit branch).

Even where it does not crash, round-tripping runtime values constant-folds the original expression (e.g. a const fetch `Foo::BAR` is baked into its value), losing source fidelity in the proxy.

## Root cause

`src/Proxy/Generator/AttributeGroupsGenerator.php:43-65` uses evaluated argument values instead of the attribute's AST.

## Proposed fix

[...]

PR #614, opened 2026-08-28
Fix enum const-expression case values, extend pointcut expressiveness, add PHP 8.5 limitations doc

### LINKED ISSUE #600: Enum proxies lose constant-expression case values (invalid backed enum generated)
## Summary

`EnumProxyGenerator` only understands literal `String_`/`Int_` case values. Any backed-enum case whose value is a constant **expression** is re-declared in the proxy enum **without a value**, producing a fatal error: a pure case inside a backed enum.

Found during the PHP 8.5 support audit (PR #597). Reproduces on PHP 8.4 / 8.5.10 / 8.6.0beta2.

## Reproduction

```php
enum Php81EnumConstExprCases: int
{
    private const int SHIFT = 2;

    case Negative  = -1;          // UnaryMinus
    case Shifted   = 1 << 2;      // BitwiseShift
    case FromConst = self::SHIFT + 10; // ClassConstFetch + Plus

    public function describe(): string { ... }
}
```

Generated proxy:

```
PHP Fatal error: Case Negative of backed enum ...\Php81EnumConstExprCases must have a value
```

Fixture: `tests/Fixtures/audit/src/Php81EnumConstExprCases.php` (audit branch).

## Root cause

`src/Proxy/EnumProxyGenerator.php` — `resolveEnumData()` (~lines 288-295):

```php
if ($stmt->expr instanceof String_)  { $caseValue = $stmt->expr->value; }
elseif ($stmt->expr instanceof Int_) { $caseValue = $stmt->expr->value; }
```

Everything else leaves `$caseValue = null`, and `EnumGenerator::addEnumCase()` then skips `setValue()`.

## Proposed fix

Stop evaluating the expression at all: carry the original `Expr` node from the enum case into `EnumGenerator` and emit it verbatim (the proxy enum lives in the same namespace, and `self::` refers to the proxy enum where the constants are preserved via the trait — note `self::SHIFT` needs the constant to be resolvable from the proxy; since class constants stay in the woven trait and trait constants are usable since PHP 8.2, `self::SHIFT` on the proxy resolves through the used trait). Evaluating via `ReflectionEnumBackedCase::getBackingValue()` is a simpler alternative (constant-fold at weave time), but emitting the AST expression preserves source fidelity; either fixes the fatal.

Add fixtures for negative, bitwise, and `self::CONST` case values.

## Environment

- goaop/framework master (4.0-dev), PHP 8.4 / 8.5.10 / 8.6.0beta2

### LINKED ISSUE #604: Pointcut expressiveness gaps for modern type/modifier system (union/DNF return types, readonly/private(set)/final)
## Summary

Two pointcut components lag behind the modern PHP type/modifier system. Neither breaks weaving — matching is just silently impossible or wrong for these constructs. From the PHP 8.5 support audit (PR #597).

## 1. `ReturnTypePointcut` cannot match union / intersection / DNF types

`src/Aop/Pointcut/ReturnTypePointcut.php:27-29` documents it:

> This implementation currently doesn't support properly matching of complex types, thus union/intersection/DNF types are not supported yet here.

Matching is a string compare + wildcard on `(string) $reflector->getReturnType()` (line ~76), so `execution(public **->*(*): Iterator)` never matches `Iterator|Generator`, and member order (`A&B` vs `B&A`) defeats exact matches.

Proposed: normalize both sides into sorted member sets (split on `|` at depth 0, `&` inside parens), and let a single-type pattern match if any union member matches; document the semantics chosen.

## 2. `ModifierPointcut` has no predicates for modern modifiers

`src/Aop/Pointcut/ModifierPointcut.php` matches only the classic `IS_PUBLIC/IS_PROTECTED/IS_PRIVATE/IS_STATIC/IS_FINAL/IS_ABSTRACT` bitmask, and `src/Aop/Pointcut/PointcutGrammar.php` only accepts `public|protected|private|final` keywords. Not expressible today:

- `readonly` properties/classes (8.1/8.2)
- asymmetric visibility `private(set)` / `protected(set)` (8.4; static asymmetric visibility in 8.5)
- hooked vs plain properties (8.4)
- `final` properties (8.4)

Proposed: extend the grammar with `readonly`, `private(set)`, `protected(set)` modifier tokens mapped onto the corresponding `ReflectionProperty` predicates (`isReadOnly()`, `isPrivateSet()`, `isProtectedSet()`, `isFinal()`), with the existing matcher-gate caveat that readonly/hooked/static properties are excluded from `access()` interception anyway (useful mainly for negative filters and class filters).

### LINKED ISSUE #605: Document PHP 8.5 support status (docs/php85-limitations.md) and fix php84-limitations.md lazy-objects claim
## Summary

From the PHP 8.5 support audit (PR #597):

1. **`docs/php85-limitations.md` does not exist.** `docs/php84-limitations.md` set the precedent; 8.5 needs the equivalent. Verified content for it (PHP 8.5.10 + 8.6.0beta2, audit harness on the PR branch):
   - **Works**: pipe operator `|>` in woven bodies; `clone with`; `#[\NoDiscard]` (copied to proxy, join-point dispatch preserves the value path); attributes on class constants incl. `#[\Deprecated]`; closures/FCC as parameter defaults (merged earlier); `final` promoted properties and static asymmetric visibility for non-intercepted properties; `self`/`parent` reflection resolution.
   - **Broken / limited** (tracked separately): closures/FCC in *attribute arguments* (#601), promoted-property interception incl. `final` promoted (#599), enum const-expression case values (#600), `new` in initializers under `INTERCEPT_INITIALIZATIONS` (#603), global constants in attribute args (#602), class-level attributes (#598).
   - Static properties (incl. 8.5 static asymmetric visibility) are never interceptable via `access()` — hooks don't exist for static properties; worth stating explicitly.

2. **Factual error in `docs/php84-limitations.md:36`**: it claims the framework's `Container` "uses lazy objects internally". It does not — `src/Core/Container.php` uses a hand-rolled closure/factory dance (`addLazyService()`/`materializeService()`). Either reword the doc or actually adopt lazy objects (#606 tracks adoption).

lisachenko:
# Summary

Four independent changes, one commit each.

## 1. Enum proxies lose constant-expression case values — Fixes #600

`EnumProxyGenerator::resolveEnumData()` only recognized `String_`/`Int_` literal case values, so `case Negative = -1;`, `case Shifted = 1 << 2;` and `case FromConst = self::SHIFT + 10;` resolved to `null`, `EnumGenerator::addEnumCase()` skipped `setValue()`, and the proxy declared a **pure case inside a backed enum** — a PHP fatal error ("Case Negative of backed enum ... must have a value") at proxy load time.

**Fix:** the parser-reflection path now carries the raw PhpParser `Expr` node through to `EnumGenerator`, which emits it verbatim (`Builder\EnumCase::setValue()` accepts `Expr` natively). `self::CONST` keeps resolving on the proxy because class constants stay in the woven trait and trait constants participate in the composing class since PHP 8.2 — **verified at runtime**, not just lint: a test weaves the fixture enum, loads the woven trait + proxy enum, and asserts `Negative->value === -1`, `Shifted->value === 4`, `FromConst->value === 12` and `from(12)` works.

Tests: `EnumGeneratorTest` (verbatim `Expr` emission), `EnumProxyGeneratorTest` (native-reflection path emits evaluated scalars, never valueless cases), `WeavingTransformerTest` (golden `php81-enum-const-expr{,-woven,-proxy}.php` fixtures + runtime load check), `EnumWeavingTest` (end-to-end project fixture `ConstExprBackedEnum`).

## 2. Pointcut expressiveness — Fixes #604

**(a) `ReturnTypePointcut` union/intersection/DNF support.** Pattern and actual type are normalized into sets of intersection groups (split on `|` at paren depth 0, then `&`; parens and leading backslashes removed; leading `?` on the *actual* type expands to `|null`). Semantics (documented in the class docblock, "not supported yet" note removed):
- single-type pattern (`string`, `Exc*`) matches if **any** member of the actual type matches (wildcards kept per member);
- composite pattern (`string|int`, `Countable&Iterator`) must correspond **one-to-one** to the actual member set, order-insensitively;
- in the pattern, `?` keeps its historical single-char-wildcard meaning (BC) — nullable patterns are written `Foo|null`.

The grammar now accepts union/intersection return types after `:` (e.g. `execution(public Example->method(*): string|int)`, `...: Countable&Iterator`, DNF paren-free as `A&B|C` which equals `(A&B)|C` per PHP precedence).

**(b) Modifier predicates `readonly`, `private(set)`, `protected(set)`** for `ModifierPointcut`/grammar, mapped to `ReflectionProperty::IS_READONLY` / `IS_PRIVATE_SET` / `IS_PROTECTED_SET`. Matching stays bitmask-based: both native reflection and `Go\ParserReflection\ReflectionProperty` expose these bits via `getModifiers()` (verified against vendor parser-reflection), so no `method_exists` guards are needed. The asymmetric tokens are lexed as single tokens (`private(set)`), keeping full BC. The LALR parse table was regenerated from the updated grammar — **zero conflicts**, all pre-existing grammar tests pass unchanged.

**Deferred (small, deliberate):** parenthesized DNF groups in the *grammar* (`(A&B)|C`) — the paren-free equivalent parses and matches identically, and `ReturnTypePointcut` itself normalizes parenthesized patterns when constructed directly; wildcards inside grammar-level return-type members (supported via direct construction). Neither requires further grammar surgery for the issue's use cases.

## 3. docs/php85-limitations.md — Fixes #605

New document modeled on `docs/php84-limitations.md`, capturing the PHP 8.5.10 / 8.6.0beta2 audit facts from PR #597: what works (pipe `|>` in woven bodies, `clone with`, `#[\NoDiscard]` propagation with join-point dispatch returning the value, attributes on class constants incl. `#[\Deprecated]`, closures/FCC parameter defaults, final promoted + static asymmetric visibility on non-intercepted properties, `self`/`parent` reflection resolution) and what is limited — each phrased as "tracked in #NNN" (#598, #599, #600, #601, #602, #603) since fixes are in flight on other branches — plus the permanent engine constraint that static properties are never interceptable via `access()` (no property hooks for statics). `docs/php84-limitations.md` untouched. README links the new page.

## 4. Hygiene — Partially addresses #610

- Deleted **12** orphaned fixtures in `tests/Instrument/Transformer/_files/` after verifying each basename is unreferenced across `tests/` and `src/`. **Note:** `yii_style.php` / `yii_style_output.php` from the issue's list of 14 are *not* orphaned — `FilterInjectorTransformerTest` still reads them — so they were kept.
- `.github/workflows/phpstan.yml`: PHP 8.4 + 8.5 matrix with per-version cache keys (`php-8.4` / `php-8.5`).
- `tests/functions.php`: typed the Symfony Finder `glob()` override as `glob(string $pattern, int $flags = 0): array` — matching how `Finder::searchInDirectory()` invokes it (string pattern, int `GLOB_*` bitmask).

Rest of #610 (PARAMETER_WIDENING removal, constructor promotion, replacement coverage fixtures) intentionally not touched — owned by other work.

# Test evidence

| Gate | Result |
|---|---|
| `php8.5 vendor/bin/phpunit` | OK — 2536 tests, 3042 assertions (1 deprecation, pre-existing on master) |
| `php8.4 vendor/bin/phpunit` | OK — 2536 tests, 3039 assertions (1 deprecation pre-existing, 1 skipped) |
| `php8.5 vendor/bin/phpstan analyze --memory-limit=1G` | OK — no errors, level 10, no new baseline entries |
| `php8.6 vendor/bin/phpunit` (informational, 8.6.0beta2) | OK — 2536 tests, 3042 assertions |

Fixes #600
Fixes #604
Fixes #605
Partially addresses #610

---


src/Aop/Pointcut/ReturnTypePointcut.php
```diff
@@ -21,41 +21,57 @@
...
     {
-        $returnTypeName = trim($returnTypeName, '\\');
+        $returnTypeName = trim($returnTypeName, " \t\\");
         if (strlen($returnTypeName) === 0) {
             throw new \InvalidArgumentException("Return type name must not be empty");
         }
-        $this->typeName = $returnTypeName;
-        $this->regexp   = '/^(' . strtr(preg_quote($this->typeName, '/'), [
-            '\\*'    => '[^\\\\]+',
-            '\\?'    => '.',
```
lisachenko, line 57: Let’s drop “?” from the regexp, it collides with PHP syntax, so “?” will not equal to one random symbol. Can be applied to all places, as a separate commit.

lisachenko: Done in d240b58 — `?` is no longer a wildcard in `ReturnTypePointcut` or `NamePointcut` (those were the only two translation sites; the grammar never lexed `?`). Bonus: with the collision gone, a leading `?` in a return-type pattern now works as a real nullable marker — `?array` matches `?array`/`array|null` returns — matching PHP syntax. Docblocks and tests updated; full suite + PHPStan green on 8.4/8.5.

---

PR #617, opened 2026-08-28
Fix attribute-class and new-in-initializer weaving fatals; typed consts, PARAMETER_WIDENING removal

### LINKED ISSUE #608: Adopt PHP 8.3 typed class constants and #[\Override] across src/
## Summary

From the PHP 8.5 support audit (PR #597); user-approved modernization batch.

## Typed class constants (8.3)

Exactly one constant in `src/` is typed (`AbstractMethodInvocation::TRAIT_ALIAS_PREFIX`). Add native types to the rest — concentrated targets:

- `src/Core/AspectContainer.php:27-67` (9 string constants)
- `src/Aop/Features.php:24-54` (int constants; see also the dead-flag cleanup in #610)
- `src/Aop/Pointcut.php:45-53` (9 int `KIND_*` constants)
- `src/Proxy/Generator/MethodGenerator.php:37-39`, `src/Proxy/Generator/ClassGenerator.php:37-39`, `src/Proxy/Generator/PropertyGenerator.php:27-34` (`FLAG_*`)
- `src/Proxy/Generator/TypeGenerator.php:41` (`BUILTIN_TYPES` → `const array`), `src/Proxy/EnumProxyGenerator.php:56,66`
- `src/Instrument/ClassLoading/CachePathManager.php:32,37`, `src/Instrument/Transformer/WeavingTransformer.php:42`

## `#[\Override]` (8.3)

Zero occurrences in `src/` despite deep hierarchies. Add it to overriding methods in framework code — transformers implementing `transform()`, `SourceTransformingLoader::filter()` (overrides `php_user_filter`), the `AbstractJoinpoint`/`AbstractInvocation`/`AbstractInterceptor` hierarchies, `Pointcut::matches()` implementations — and replace bare `@inheritDoc` docblocks where the attribute is the enforced version.

**Caveat**: never add `#[\Override]` to test-project fixture classes that get woven — the weaver strips it from intercepted methods for trait-alias reasons (`stripOverrideAttributeFromInterceptedMethods()`), and fixtures are meant to exercise that path deliberately, not accidentally.

### LINKED ISSUE #615: Weaving an attribute class fatals: #[\Attribute] / #[\AllowDynamicProperties] are invalid on the woven trait
## Summary

Follow-up to #598 (fixed in #613). Class-level attributes are now correctly preserved on the woven trait — but two built-in attributes are **compile-time rejected on traits**, so weaving a class that carries them is a new fatal:

```
PHP Fatal error: Cannot apply #[\Attribute] to trait Ns\ConstAttr__AopProxied
```

Verified on PHP 8.5.10: `#[\Attribute] trait T {}` and `#[\AllowDynamicProperties] trait T {}` are both compile fatals (`#[\Deprecated]` on a trait is fine). So any **attribute class** (or `#[AllowDynamicProperties]`-marked class) inside the weaver's include paths breaks at load — audit fixtures `tests/Stubs/Audit/{ConstAttr,ExprAttr,RichAttr}.php` reproduce it (audit branch PR #597, harness `Php85AuditScratchTest`).

## Proposed fix

In `WeavingTransformer`, when converting a class to a trait, drop attribute groups whose (FQ-resolved) name is `Attribute` or `AllowDynamicProperties` from the **trait** while keeping them on the generated **proxy class** (which is a real class and re-declares attributes via `AttributeGroupsGenerator`) — same split already used for the `readonly` class modifier. All other class-level attributes stay on the trait as today.

An alternative is to skip weaving classes marked `#[\Attribute]` entirely (attribute classes are rarely useful join-point targets), but the drop-from-trait approach keeps behavior uniform.

## Environment

- goaop/framework master @ c52e072 (after #613), PHP 8.4 / 8.5.10 / 8.6.0beta2

### LINKED ISSUE #616: Intercepted promoted property with `new`-in-initializer default emits invalid hook property default
## Summary

Follow-up to #599 (fixed in #613). The demotion approach correctly keeps a promoted parameter's default in the demoted constructor parameter — but the **proxy hook property** now copies that default verbatim. For a PHP 8.1 `new`-in-initializer default that's invalid: `new` is allowed in parameter defaults but **not** in property initializers.

## Reproduction

```php
class Php81NewInInitializers
{
    public function __construct(
        private Collaborator $service = new Collaborator('ctor-default'),
    ) {}
}
```

With a property advice on `$service`, the proxy contains:

```php
private \Ns\Collaborator $service = new Collaborator('ctor-default') {
    get { ... }
```

```
PHP Fatal error: New expressions are not supported in this context
```

Reproduced by audit fixture `tests/Stubs/Audit/Php81NewInInitializers.php` (audit branch PR #597).

## Proposed fix

In `AbstractInterceptedPropertyGenerator` (the default-from-`Param` logic added by #613): only carry the default onto the hook property when the expression is a valid property initializer — i.e. contains no `New_` node (a small AST check; `new` is the only param-default construct excluded from property initializers). When skipped, the property stays uninitialized and still gets its value through the injected constructor assignment (`$this->service = $service;`), which the demotion already generates; the existing `isInitialized()` guard in the get hook handles the pre-construction window.

## Environment

- goaop/framework master @ c52e072 (after #613), PHP 8.4 / 8.5.10 / 8.6.0beta2

lisachenko:
# Summary

Two confirmed bug fixes plus the approved modernization batch, in four commits (one per task). All gates green on PHP 8.4 and 8.5 (phpunit + PHPStan level 10 with no new baseline entries); the full suite also passes on PHP 8.6.0beta2 (informational).

## 1. Fixes #615 — weaving an attribute class fatals

Class-level attributes survive class→trait conversion since #613, but `#[\Attribute]` and `#[\AllowDynamicProperties]` are compile-time invalid on traits ("Cannot apply #[\Attribute] to trait"), so weaving any attribute class was a fatal at woven-trait load.

`WeavingTransformer` now removes exactly these attribute entries from the woven trait tokens (resolved names from the class node's `attrGroups`):
- a group consisting only of incompatible attributes is blanked out entirely, preserving contained newlines (line numbers stay intact for XDebug);
- in a grouped attribute (`#[\Attribute, SomethingElse]`) only the incompatible entry plus one adjacent comma is removed.

The proxy class still copies the original attribute groups from the AST (`AttributeGroupsGenerator`), so reflection on the proxied attribute class keeps reporting `#[\Attribute(...)]` with original arguments — asserted in the new test.

**Tests:** new golden fixture `php80-attribute-class.php`/`-woven.php` (attribute class with args, grouped attribute, `#[\AllowDynamicProperties]`), updated `php80-class-attribute-woven.php` golden, proxy attribute assertion.

## 2. Fixes #616 — new-in-initializer default copied onto hook property

An intercepted promoted property like `private Collaborator $service = new Collaborator('x')` produced a proxy hook property carrying the `new` default — a compile error ("New expressions are not supported in this context"): `new` is legal in parameter defaults but illegal in property initializers.

`AbstractInterceptedPropertyGenerator` now skips the AST default when it contains any `PhpParser\Node\Expr\New_` (NodeFinder). The hook property stays uninitialized; the demoted constructor parameter keeps the default, the injected `$this->prop = $prop;` assignment routes the value through the set hook, and `hasPotentiallyUninitializedTypedProperty()` now reports true so the `isInitialized()` guards cover the pre-construction window.

**Tests:** targeted `WeavingTransformerTest` assertions (woven + proxy content, proxy parseability) with fixture `php81-new-in-initializer.php`, plus a functional runtime test (`ClassWeavingTest::testNewInInitializerPromotedPropertyWeaving`) that weaves `NewInInitializerClass` in a subprocess, instantiates it without arguments and confirms the constructor default materializes (`['seed']`). Verified to fail before the fix.

## 3. Fixes #608 — typed class constants + `#[\Override]`

- Every class/interface constant in `src/` outside the #614-reserved files now has a native type: `AspectContainer` prefixes, `Features` flags, `Pointcut::KIND_*`, generator `FLAG_*`/`VISIBILITY_*`, `TypeGenerator::BUILTIN_TYPES` (array), `CachePathManager` cache file names, stream-filter identifiers and cache suffixes.
- `#[\Override]` added to ~110 overriding methods across `src/` (verified exhaustively via reflection): transformer `transform()` implementations, `SourceTransformingLoader::filter()`, the `AbstractJoinpoint`/`AbstractInvocation`/`AbstractInterceptor` hierarchies, `Pointcut::matches()`/`getKind()` implementations, container/aspect-loader extensions, console commands, proxy generators, node visitors. Docblocks that were only `{@inheritDoc}` were replaced by the attribute; docblocks adding `@return`/`@throws` info were kept. No `#[\Override]` added under `tests/Fixtures/` or `tests/Stubs/` (the weaver's Override-stripping paths exercise those deliberately).

## 4. Partially addresses #610 — with #614 covering the rest

- **[BC BREAK] `Features::PARAMETER_WIDENING` removed** together with the whole `$useParameterWidening`/`$useTypeWidening`/`$useWidening` plumbing through `WeavingTransformer`, all four proxy generators, `InterceptedMethodGenerator`, `InterceptedConstructorGenerator`, `FunctionParameterList` and `MethodGenerator`/`FunctionGenerator`/`ParameterGenerator::fromReflection()`. The flag has been a no-op concern since the PHP 7.2 baseline; generated proxies always keep original parameter types. CHANGELOG entry added under 4.0.0 (unreleased). Kernel configurations passing the flag must drop it.
- **Constructor property promotion** for declare-then-assign constructors in `src/Instrument/` (`BaseSourceTransformer`, `Enumerator`, `CachePathManager`, `AopComposerLoader`, `WeavingTransformer`) and `src/Proxy/Generator/` (`ClassGenerator`, `TraitGenerator`, `PropertyGenerator`, `ParameterGenerator`, `ValueGenerator`, `DocBlockGenerator`, `MethodGenerator`, `FunctionGenerator`), `readonly` where never reassigned. Left as-is: properties fed through `array_values()` or derived from other constructor arguments (`$container`, `$options`-derived fields), and `FileGenerator` (no constructor).
- **Restored golden coverage of general PHP 8.0–8.3 syntax** through the current weaver: new fixture `php80-82-syntax.php` + `-woven.php` + `-proxy.php` covering constructor promotion (non-intercepted property), new-in-initializer parameter default, named arguments, `match`, nullsafe operator, enum usage in a method body, readonly property, first-class callable and a typed class constant; wired into `WeavingTransformerTest`. No fixtures were deleted.

## Deferred until #614 merges (do-not-touch in this PR)

- `EnumProxyGenerator` / `EnumGenerator`: no typed-const or `#[\Override]` changes. Exception: the PARAMETER_WIDENING removal required a **minimal** edit to `EnumProxyGenerator`'s constructor (drop the 4th parameter, its docblock line and the property assignment — 7 lines); small conflict risk with #614 accepted and flagged here.
- `ReturnTypePointcut`, `ModifierPointcut`, `PointcutGrammar`, `PointcutParseTable`: constants/overrides there remain untyped/unattributed for a follow-up.
- `.github/workflows/phpstan.yml` 8.5 job, `tests/functions.php` typing and orphaned-fixture deletions: carried by #614.
- Still open after both PRs: the static-singleton/`extract()` cleanup from #610 item 4.

## BC note

`Features::PARAMETER_WIDENING` (int 128) no longer exists and the proxy-generator constructors / `fromReflection()` factories lost their widening parameter. Anyone constructing these classes directly or passing the feature flag must update (4.0 changelog line included).

## Test evidence

- `php8.5 vendor/bin/phpunit` — OK (2524 tests, 3055 assertions)
- `php8.4 vendor/bin/phpunit` — OK (2524 tests, 3044 assertions, 4 skipped: 8.5-only)
- `php8.5 vendor/bin/phpstan analyze --memory-limit=1G` — level 10, no errors, baseline unchanged (3 entries)
- `php8.6 vendor/bin/phpunit` (informational, 8.6.0beta2) — OK (2524 tests)
- Regression checks: the new #616 functional test and the #615 golden test both fail on master before their fixes.

---


src/Aop/Framework/AbstractInterceptor.php
```diff
@@ -63,6 +63,7 @@ public function __construct(
         protected readonly string $pointcutExpression = ''
     ) {}
 
+    #[\Override]
```
lisachenko: What the point of adding Override everywhere? I would like to have only necessary set of such attributes where needed. Or remove them all entirely

lisachenko: Fair — the attribute is compile-time intent-checking only, and 110 of them is a lot of churn for that. Removed them all in 2599eba (typed constants stay; the weaver's Override-*stripping* logic and its docs are untouched, since that handles user code). Gates re-verified green on 8.4/8.5. I'll note on #608 that the Override half is declined so the issue closes on the constants alone.

---

PR #620, opened 2026-08-29
feat(aop)!: first-class callable advices

lisachenko:
## What

Brings **first-class callable advices** into the framework as the main way advices are woven into generated code (supersedes the `feature/direct-first-class-advices` WIP branch, rebased onto `master` and reorganized into logical commits).

Generated proxies now reference each advice as a closure created with first-class callable syntax directly on the aspect instance:

```php
static $__joinPoint = InterceptorInjector::forMethod(
    self::class,
    'doSomething',
    [
        Interceptor::before(The::aspect(MonitorAspect::class)->beforeMethodExecution(...)),
    ],
    $this->__aop__doSomething(...),
);
```

Advices registered in the container as plain closures (not aspect methods) are resolved lazily through the new `The::advice('advisorId')` accessor, which unwraps `Advisor`/interceptor values down to the raw advice closure.

## Commits

1. **feat(aop): generate first-class advice callables** — the proxy generation switch: `The`/`Interceptor` facades, `GeneratedInterceptor` descriptors, `InterceptorListGenerator`, removal of `LazyAdvisorAccessor` (rebased onto current master, adapting master's newer fixtures/tests).
2. **feat(aop)!: replace advice marker interfaces with AdviceTypeEnum** — drops `AdviceBefore`/`AdviceAfter`/`AdviceAround`; `Advice::getType(): AdviceTypeEnum` now carries the advice kind and its sorting priority.
3. **feat(aop)!: require aspect advice methods to be public** — FCC wiring calls advice methods on the aspect instance from generated code, so the attribute loader rejects non-public advice methods with an `AspectException` (methods with only `#[Pointcut]` keep any visibility).
4. **feat(aop): resolve closure advices through The::advice() accessor** — container-backed closure advices become weavable; joinpoint flattening rejects non-`Advice` advisor values loudly; `GeneratedInterceptor::fromAdvisorId()` removed.
5. **docs: document first-class callable advices** — CHANGELOG 4.0.0 entries and README rework presenting FCC advices as the main approach.
6. **docs: refresh AGENTS.md for first-class callable advice wiring** — agent guides updated to the new proxy dispatch and advice-wiring components.

## BC breaks

- Aspect advice methods (`#[Before]`, `#[After]`, `#[Around]`, `#[AfterThrowing]`) must be **public**.
- `AdviceBefore`/`AdviceAfter`/`AdviceAround` marker interfaces removed; every `Advice` implementation must implement `getType(): AdviceTypeEnum`.

## Testing

- `./vendor/bin/phpunit` — 2609 tests, 3216 assertions, green (validated on every commit of the series).
- `./vendor/bin/phpstan analyze --memory-limit=512M` — level 10, no errors on every commit.


src/Proxy/FunctionProxyGenerator.php
```diff
@@ -56,7 +57,14 @@ public function __construct(
         $this->fileGenerator = new FileGenerator();
         $this->fileGenerator->setNamespace($namespace->getName());
         $this->fileGenerator->addUse('Go\Aop\Framework\InterceptorInjector');
+        $this->fileGenerator->addUse('Go\Aop\Framework\Interceptor');
```
lisachenko: use ::class notation here, eg Interceptor::class

src/Proxy/FunctionProxyGenerator.php
```diff
@@ -101,9 +109,7 @@ protected function getJoinpointInvocationBody(ReflectionFunction $function): str
         }
 
         $functionAdvices = $this->adviceNames[AspectContainer::FUNCTION_PREFIX][$function->name];
-        $advicesArray    = new ValueGenerator($functionAdvices);
-        $advicesArray->setArrayDepth(1);
-        $advicesCode = $advicesArray->generate();
+        $advicesCode = (new InterceptorListGenerator(array_values($functionAdvices)))->generate('            ');
```
lisachenko: Why there is such a strange empty string inside generate call?

lisachenko: here and in other places

src/Proxy/FunctionProxyGenerator.php
lisachenko: Done in e6fc8f2 — all 25 string-literal `addUse()` calls across the four proxy generators now use `::class` references, with the corresponding imports added.

---

src/Proxy/FunctionProxyGenerator.php
lisachenko: It was 12 spaces of continuation-line indentation for the rendered interceptor list (the list is re-indented to sit inside the generated `InterceptorInjector::for*()` call). Every call site passed the same value, so in e6fc8f2 it became a documented `JOINPOINT_ARGUMENT_INDENT` constant used as the default of `generate()`, and all six call sites (this one and the other generators) now just call `generate()` with no argument.

---

